PR #5, opened 2017-01-21
Fix title CSS

hallcristobal:


CryZe:
Can you remove the livesplit_core.js change from this? Just changing CSS shouldn't require a modification of livesplit_core.js.

hallcristobal:
updated

PR #12, opened 2017-06-02
Improve Run Editor look and feel

hallcristobal:
Improve the look and feel of the Run Editor as well as update to the new LiveSplit Icon


CryZe:
👌

PR #29, opened 2017-06-18
Electron Standalone App

hallcristobal:


hallcristobal:
Ideally, yes. However we need to work out a lot of things (obviously). Getting everything working in it's entirety is the the first priority, but because there are some functions that are not supported with electron vs browser as well as some added functionalities from using Electron it will take us some time to decide how to go about this.

Tagadda:
Huh, why closing this PR ?

CryZe:
We actually just revived the Electron build yesterday. However this branch here was too outdated, so I assume we'll just have a new PR.

PR #51, opened 2017-10-29
Integrate FontAwesome into the repo.

TheEssenceOfDarkness:
This pull request in a nutshell:
* Created assets directory.
* Moved ./images into ./assets (while updating anything pointing to ./images)
* Added FontAwesome into the repo to remove a dependency on having it loaded externally.


CryZe:
There's a src/css folder with css and fonts in there, can you make sure we don't have these in 2 locations?

TheEssenceOfDarkness:
I fixed it. Also, I have no idea how I closed this.

TheEssenceOfDarkness:
That should not have errored when running webpack. It worked on my end.

CryZe:
Your PR has no bug. They just released TypeScript 2.6 and it seems like they removed some unsoundness in the type system which makes it error out now. I'll fix it on master.

CryZe:
I fixed the TypeScript bug. Can you rebase your PRs? I'll merge them then.

TheEssenceOfDarkness:
Do I need to rebase the new commit on master in? They should be able to merge without it.

CryZe:
That might work too.

PR #54, opened 2017-11-01
Added basic usage to readme.md

TheEssenceOfDarkness:
There is no documentation on the hotkeys, so it makes sense to add some. I also added "*.swp" to the .gitignore as my text editor (vim) creates a swap file when a file is opened.

I am debating if I should put this in a seperate file.


CryZe:
What happened here?

PR #66, opened 2018-02-18
LiveSplit One, now at 60 FPS.

### LINKED ISSUE #64: Figure out how to stabilize the framerate more
It's not really bad, but I think we should be using requestAnimationFrame() or something, but last time I tried it, it was super janky, so we should look into this more.

TheEssenceOfDarkness:
I shouldn't even have to put a description here, just look at the screenshot and the diff.

screenshot_2018-02-18_14-50-16

On lower end machines it will drop to 30, but it is otherwise pretty stable on most hardware. If I wanted to, I could have went higher, but going any higher would have been overkill.

### DIFF
--- a/src/layout/AutoRefreshLayout.tsx
+++ b/src/layout/AutoRefreshLayout.tsx
@@ -28,7 +28,7 @@ export default class AutoRefreshLayout extends React.Component<Props, State> {
                     layoutState: this.props.getState(),
                 });
             },
-            1000 / 30,
+            1000 / 60,
         );
     }
 
--- a/src/layout/DragAutoRefreshLayout.tsx
+++ b/src/layout/DragAutoRefreshLayout.tsx
@@ -39,7 +39,7 @@ export default class AutoRefreshLayout extends React.Component<Props, State> {
                     layoutState: this.props.getState(),
                 });
             },
-            1000 / 30,
+            1000 / 60,
         );
     }
 
--- a/src/ui/SideBarContent.tsx
+++ b/src/ui/SideBarContent.tsx
@@ -49,7 +49,7 @@ export class SideBarContent extends React.Component<Props, State> {
     public componentWillMount() {
         this.intervalID = setInterval(
             () => this.update(),
-            1000 / 30,
+            1000 / 60,
         );
     }
 


CryZe:
Alright, so I haven't quite looked at this now, but I can't imagine this actually being the solution. The issue wasn't about increasing the frame rate to 60 FPS, but instead to stabilize the frame rate. At its current configuration, I had a ton of dropped frames here and there and inconsistencies of around 28 - 32 FPS. So I don't quite see how just changing it to 60 FPS supposedly fixes that.

It does seem like it is very stable in your image, but you also don't have a lot going on on your layout, so I'd suggest at least having like 10 splits with icons and a graph.

Also like I said, this is about stabilizing the framerate, not necessarily upgrading to 60 FPS. The thing about 60 FPS is that it's definitely smoother, but at the same time requires at least twice the processing resources, so both your GPU and CPU are going to be used at least twice as much, so that's double the cpu usage (which we want to keep low), and also double the amount of battery drain on mobile devices. I don't quite think that the slightly smoother animation is worth using up twice the amount of energy. I'd rather have a stable 30 FPS with no dropped frames than a huge amount of wasted cpu performance and energy.

TheEssenceOfDarkness:
Alright, so in practice the cpu usage is 3.2x as much (not good), so we should hold off on this change until it can be brought down, and the Samsung Galaxy S7 I used for testing on mobile gets 30fps anyways. I will be closing this for now, but when I can get the cpu usage lower, I will reopen it.

TheEssenceOfDarkness:
I am retarded. The performance issues were caused by a dev build instead of a production build. Anyways, here is the performance of 10 splits with icons, and a graph.
screenshot_2018-02-19_20-02-34

CryZe:
Well, while it does some a bit more stable than when I originally tested it, it doesn't solve the battery drain issue at all. Cause no matter how fast it is, it will always at least drain the battery twice as much as 30 FPS. So no matter how much we actually drain the battery, draining it half as much is pretty much always the preferable choice compared to having it run slightly more fluid. This is something that usually gets ignored completely in the 30 FPS / 60 FPS discussions, but is very relevant here, as the web version is likely to be used on phones. So while I think 60 FPS does make sense, it should definitely be something that you can toggle.

TheEssenceOfDarkness:
I my opinion, we should hold off on this entirely until we can make the code for rendering the layout more efficent, since right now we are having React rerender the whole layout every frame, which is very inefficent.

However when we get the settings menu this can be made a toggle.

CryZe:
I don't quite think you get it though. The code is already really fast and could easily render like 500 FPS or so. So while React definitely isn't the most efficient renderer for this, and there being lots of other ways to improve this too, none of this matters much as it will always be true that 30 FPS is just sooo much more battery efficient than 60 FPS, no matter what the frame time actually is. So yes, we should improve the frame times, but it will never really be the case that 60 FPS makes sense for all platforms. I think it should definitely be a setting which should be 30 FPS by default and maybe 60 FPS by default on normal PCs.

TheEssenceOfDarkness:
My main reason for deciding to hold off on it was mobile devices not being able to matain 60 fps in the first place.

PR #68, opened 2018-02-19
Match LiveSplit's saving behavior.

### LINKED ISSUE #11: Reset before Saving

### LINKED ISSUE #71: Save / Export does not save split times
#### Steps to reproduce the issue

1.  Open https://livesplit.github.io/LiveSplitOne, I recommend doing this in a private browser window so your localStorage does not get overwritten
2. Finish a run, even if it's only the default run with one split so you have split times
3. Save or export the file file
4. Refresh the tab
5. Import your savefile

#### What's the expected result?

- On refreshing the tab in (4.) I expect the split times to be included so I can compare my next run to them
- Same goes for the exported xml in (5.)
- When I download a sample xml from splits.io the split times are actually included in the xml file

#### What's the actual result?

- On refreshing the tab in (4.) the restored xml in the localStorage does not include the split times generated in (2.)
- Same goes for the exported xml in (5.)

#### Additional details / screenshot

This might be an issue with livesplit-core instead of LiveSplitOne

TheEssenceOfDarkness:
Closes #11 
Closes #71 

Timer now resets before saving splits.

### DIFF
--- a/src/ui/LiveSplit.tsx
+++ b/src/ui/LiveSplit.tsx
@@ -245,7 +245,7 @@ export class LiveSplit extends React.Component<{}, State> {
     }
 
     public saveSplits() {
-        const lss = this.state.timer.readWith((t) => t.getRun().saveAsLss());
+        const lss = this.state.timer.readWith((t) => t.saveAsLss());
         localStorage.setItem("splits", lss);
     }
 


CryZe:
This is not the solution, as you might want to save mid-run, but don't actually want to reset it. Therefore you kind of want to figure out if there's an active attempt, and if there is, clone it and reset that one.

TheEssenceOfDarkness:
I didnt know exactly how you wanted it implemented since the issue had no description. How you feel about having it reset only if the run has ended?

CryZe:
Well, that's not how the original LiveSplit behaves. If you save mid-run it should store all the information of the current attempt as if you ended your current attempt at this point in time. But all of this should happen without the user's actual current attempt forcibly being reset.

TheEssenceOfDarkness:
Could this be better handled in livesplit core before I start writing the code for it?

CryZe:
I think it could make sense to have it in lsc, but it's also just two lines of code, allowing you to also somewhat reuse the copy in some way, so I think it's fine if just LSO does this.

src/ui/LiveSplit.tsx
```diff
@@ -245,6 +245,7 @@ export class LiveSplit extends React.Component<{}, State> {
     }
 
     public saveSplits() {
+        this.state.timer.writeWith((t) => t.reset(true));
         const lss = this.state.timer.readWith((t) => t.getRun().saveAsLss());
```
CryZe: Also as an additional hint, you are locking the timer twice here (writeWith, readWith). This is fine, since they are separate, so it's not a deadlock, however for the very short period of time in between, something else could lock the timer and completely modify its state, meaning that it may not actually be properly reset anymore. So you actually want to combine these into just a single writeWith.

TheEssenceOfDarkness: I will keep this in mind when I get around to making this match livesplit's behavior

TheEssenceOfDarkness:
Alright, my current plan to rewrite this to match LiveSplit's behavior is:

* Lock the timer state with a `writeWith`

* Clone the splits and current attempt for restoring later

* End the current attempt 

* Save the splits

* Restore the Splits and Current attempt from the clone

* Clean everything up

* Unlock the timer

Does this sound like it might work?

CryZe:
Alright, looking at it again, you definitely want to reset the clone, but that's not really possible cleanly at the moment, as you would want to fully clone the timer object for that, which is not a thing in at least the TypeScript / C API. So this is definitely blocked in some fashion by livesplit-core.

hallcristobal:
I almost feel like this would be better to put in lsc, rather than LSO

CryZe:
There's 3 options:
1. expose timer.clone() and do all the resetting + saving logic in LSO
2. have a timer.getResettedRun(), which is a clone of the current run, but resetted
3. have a timer.saveAsLss() which does all of this including the saving

Also resetted isn't even a word, so it would be getResetRun or something, which kinda sounds bad.
Option 3 has the disadvantage that it only does saving as lss, so it's not flexible.
Option 1 forces this logic into the UI, which is suboptimal, as it's easy to make this mistake of not actually resetting your current attempt properly.

hallcristobal:
Is there way to pass in what kind of splits your saving to it and then have lsc handle it that way?

TheEssenceOfDarkness:
I really don't like option 3, due to it only being useful for one purpose. I would go with either 1 or 2 simply because they both provide some level of flexibility for other developers using the library. In this specific case however, 2 seems like the best option for this. 

If you do use option 2, the function should be named `getResetRun()`.

CryZe:
I kinda settled for both Option 1 (which is already a thing and not exposed) and Option 3. So for saving specifically, being a common use case, we have a nice method that does everything correctly for you, and if you want to do something that is a slightly different use case, then you can very easily emulate Option 2, by more or less just using `timer.clone().reset(true).getRun()` (ignoring a few tiny details).

hallcristobal:
My main argument was that it was easy for a user (or developer) to incorrectly implement saving & reset. And since it's, approximately, the same process it would make sense to me to either make it one function, or make it require the correct functions to be called

TheEssenceOfDarkness:
I will go ahead and make a commit for option 3 (but not push it as to avoid failing the CI check). What function name did you settle on for it?

CryZe:
It would just be `t.saveAsLss()` without the .getRun()

CryZe:
Alright, I just implemented it: https://github.com/LiveSplit/livesplit-core/commit/b863ecee39945ba32b4a6bfc457b60c8b8a0164f

TheEssenceOfDarkness:
Now all we need to do is update livesplit-core in this repo, and we will be good to go.

CryZe:
You can do that very easily by just checking out the commit I link in the livesplit-core submodule.

TheEssenceOfDarkness:
I just rebased to

CryZe:
Was there anything blocking this? I don't quite remember and this has apparently been sitting here for a while.

CryZe:
Looks fine regardless, so I'll just merge this :D

PR #72, opened 2018-04-28
Correct Run Editor CSS

hallcristobal:


CryZe:
The Layout Editor regressed with this:
Now:

Before:


Other things may have regressed too. I feel like we need to fully split these into different CSS with only the basic styling as base rules for both.

CryZe:
@hallcristobal Are you still working on this?

hallcristobal:
I unfortunately completely forgot about this.
Gonna finish now and commit.

But while I'm remembering it, how about moving to Scss?
My thought is that then we can have, say,
```SCSS
#run-editor {
    .table {...}
    thead {...}
}
#layout-editor {
    .table {...}
    thead {....}
}
.component {
    .timer {...}
    .splits {...}
}
```

PR #81, opened 2018-05-10
Correct Title Component Layout

### LINKED ISSUE #80: Centered Title with Game Icon isn't really centered
The name of the game is too far right: 

I'm not really sure if the second line is correct either.

hallcristobal:
Correct Center text when Image is visible
Correct Single line vertical Alignment

Closes #80

### DIFF
--- a/src/layout/Title.tsx
+++ b/src/layout/Title.tsx
@@ -7,11 +7,32 @@ export interface Props { state: LiveSplit.TitleComponentStateJson }
 
 export default class Title extends React.Component<Props> {
     private iconUrl: string;
+    private gameNameElement: React.RefObject<HTMLElement>;
 
     constructor(props: Props) {
         super(props);
 
         this.iconUrl = "";
+        this.gameNameElement = React.createRef();
+    }
+
+    public componentDidMount() {
+        const container = this.gameNameElement.current;
+        if (container === null) {
+            return;
+        }
+        if (this.getIconUrl() !== ""
+            && container.classList.contains("justify-center")
+            && container.parentElement != null
+            && container.offsetWidth > container.parentElement.offsetWidth - 100) {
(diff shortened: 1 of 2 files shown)


src/layout/Title.tsx
```diff
@@ -14,6 +14,25 @@ export default class Title extends React.Component<Props> {
         this.iconUrl = "";
     }
 
+    public componentDidMount() {
+        const container = document.getElementsByClassName("title-game")[0] as HTMLElement;
```
CryZe: Uhm, why are we circumventing React here?

hallcristobal: I wasn't sure how to get the width of a rendered element without putting code to execute immediately after render. If there is a way and you know it I'm all for it.

CryZe: I feel like this could break if there's multiple title components involved though. I'm pretty sure React has something that gives you the proper element directly.

src/layout/Title.tsx
```diff
@@ -14,6 +14,25 @@ export default class Title extends React.Component<Props> {
...
+        if (container === null || container === undefined) {
```
CryZe: if (container == null) should be enough

src/layout/Title.tsx
CryZe: Other than that, I guess I'm fine with it, if this is really necessary.

hallcristobal: Well, could store it in the state that would allow us to access it directly, but with how the state onliest reacts with lsc I want sure if it was the best idea

hallcristobal: Also, implemented this because if there is an image and the text is sent to center, if the text is too long it will go under the icon, this circumvents this by checking if it's long at render so and then sets it to size to the right of the image.

PR #134, opened 2018-12-31
Add custom scrollbars for WebKit based browsers.

TheEssenceOfDarkness:
This should look a lot nicer on WebKit based browsers.

screenshot_2018-12-31_12-47-11

I also had to make sidebar 10px wider to account for the new scrollbar. Otherwise, we'd get stuff like this:

screenshot_2018-12-31_13-08-57


src/css/livesplit.css
```diff
@@ -117,7 +117,7 @@ body {
     background: hsl(0, 0%, 10%);
     padding: 10px;
     padding-top: 15px;
-    width: 250px;
+    width: 260px;
```
CryZe: Is that necessary? The sidebar is already kinda wide and I'd like to keep it as little as possible.

TheEssenceOfDarkness: Unfortunately, yes. Otherwise you'd end up with the second screenshot.

PR #168, opened 2019-09-01
Make minor fixes to the webkit scrollbars.

TheEssenceOfDarkness:
Noticed a couple of minor tweaks could be made here.

### DIFF
--- a/src/css/main.scss
+++ b/src/css/main.scss
@@ -99,6 +99,7 @@ button.disabled:active {
 /* WebKit Scrollbars */
 ::-webkit-scrollbar {
   width: 10px;
+  height: 10px;
 }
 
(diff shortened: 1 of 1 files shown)


TheEssenceOfDarkness:
I just resolved a merge conflict due to the recent change to split up the CSS so this should be good to merge now.

CryZe:
If there's even anything that can be done about this.

TheEssenceOfDarkness:
>I'm getting some weird white square in the bottom right when both scrollbars are visible.

I probably forgot to copy that over when fixing the merge conflict. I have no idea how that happened.

CryZe:
Well there's still a visible square:

PR #169, opened 2019-09-02
Prefer requestAnimationFrame over setInterval

### LINKED ISSUE #64: Figure out how to stabilize the framerate more

DarkRTA:
`requestAnimationFrame` is preferred over `setInterval` for drawing frames as it allows the browser itself to make decisions on when to draw the layout. Unfortunately, we are no longer capped at 30fps which will result in higher CPU usage, but I feel the tradeoff is acceptable here as the browser will throttle itself if needed.

On the plus side however, input latency will be lower overall and frametimes will be more consistent avoiding stuff as seen in #64.

Closes #64.

### DIFF
--- a/src/layout/AutoRefreshLayout.tsx
+++ b/src/layout/AutoRefreshLayout.tsx
@@ -14,7 +14,7 @@ export interface State {
 }
 
 export default class AutoRefreshLayout extends React.Component<Props, State> {
-    private intervalID: any;
+    private reqId: any;
 
     constructor(props: Props) {
         super(props);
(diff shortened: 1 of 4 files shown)


CryZe:
I'm honestly not sure what to do with this. The low CPU usage is super important to me. I think we should gate this behind a setting that the user can set. "Reduce Battery Usage [x]" or so

wooferzfg:
Not sure how stable the frame rate is with this, but it's possible to use `requestAnimationFrame` while capping the FPS: https://stackoverflow.com/a/19772220

wooferzfg:
I still get a pretty unstable frame rate with this branch:

On my 144 Hz monitor:
chrome_OqmkPGG9he

On my 60 Hz monitor:
chrome_3hqV6SisM2

wooferzfg:
I don't know if this is steady enough for what we want to accomplish:

chrome_tsLBF0l4nJ

Even if we give up on using `requestAnimationFrame`, I definitely want to keep the extracted `AutoRefresh` component because it gets rid of code duplication in 3 places.

CryZe:
Isn't this super unstable with refresh rates that are not a multiple of 30?

DarkRTA:
To be honest I really do not care *too* much if this gets merged but I want to mention that this was perfectly stable on my setup. Not like this is a big deal to begin with as we are mostly just updating some text.

I do agree that we definitely want to keep the extracted `AutoRefesh` component for the sake of code quality though.

wooferzfg:
The stability for me is about the same on my 144 Hz monitor as it is on my 60 Hz monitor.

CryZe:
I'll merge it then. We can do a followup PR for a "60 FPS" setting if we want that.

PR #179, opened 2019-10-23
Fix styling on Splits component

### LINKED ISSUE #153: Current Split Gradient is behind Split Background
Not sure how I never noticed this, but the current split gradient is being rendered behind the split's background. Normally the split's background is mostly transparent, so I guess that's how this happened.

wooferzfg:
- Fix #153 by switching which background we use for the inner style and the outer style.
- Fix the background not extending fully horizontally on empty splits. You can easily reproduce this issue by just setting a background color with the default splits and layout. I fixed this by passing `maxColumns` as a prop to `Split`, but it would probably be easier if `livesplit-core` did this instead.

### DIFF
--- a/src/css/Splits.scss
+++ b/src/css/Splits.scss
@@ -33,6 +33,7 @@
       .split-icon {
         max-width: 18px;
         height: 18px;
+        object-fit: contain;
         vertical-align: middle;
         filter: drop-shadow(2px 2px 1px rgba(0, 0, 0, 0.3));
       }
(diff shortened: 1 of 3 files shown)


wooferzfg:
I also fixed some issues with how images are positioned and scaled. Here's the splits file I used to test it: [SM_Low_Ice_Fixed.zip](https://github.com/LiveSplit/LiveSplitOne/files/3770134/SM_Low_Ice_Fixed.zip)

Before:
chrome_yJlmH4ava1

After:
chrome_D5yXVj2kYE

PR #180, opened 2019-10-23
Indent split names if at least one split has an icon

### LINKED ISSUE #152: Remove / show icons column based on has_icons
It was added in this commit: https://github.com/LiveSplit/livesplit-core/commit/c0f34793bf204a0a691f635d82cd0de8a8776b6d

wooferzfg:
Fix #152


CryZe:
Seems to work well :)

PR #181, opened 2019-10-26
Add support for icon and segment name in detailed timer

### LINKED ISSUE #37: Implement Detailed Timer Icon and Segment Name
livesplit-core supports this now.

wooferzfg:
I tried my best to handle different heights and widths, but it's hard with CSS.

Fixes #37 

I also noticed a difference between livesplit-core and the original LS - after the final split, LS still shows the segment name, but livesplit-core doesn't. In my opinion, we should be showing the last segment's name and icon after the run ends because we're still showing the comparisons for that segment.

LS: 
Code_PS0KuxhnSN

LSO with icon and segment name enabled:
chrome_k0j7VHm7nU

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 node_modules
 dist
 npm-debug.log
-src/livesplit_core.js
 src/livesplit_core.wasm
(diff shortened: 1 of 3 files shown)


CryZe:
I'll fix that as part of this PR I guess.

PR #182, opened 2019-10-26
Fix split labels

wooferzfg:
I think the issue was caused by a mismatching number of divs between the split labels and the regular splits.


CryZe:
I'm now getting word wrapping with this branch

PR #184, opened 2019-10-28
Fix timer shadow not working in Firefox

### LINKED ISSUE #142: Timer text shadow not rendered in Firefox
With the same layout, in Firefox 

screenshot from 2019-01-25 15-26-53

in Chrome
screenshot from 2019-01-25 15-27-21

The [docs to  demonstrate `text-shadow`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-shadow) render correctly for me in both browsers, and there shouldn't need to be a `-moz-text-shadow`.

Changing the alpha to `1` or changing any part of the `text-shadow` expression doesn't show a shadow in FF either. 

It's not a big deal since a userscript can clear the `text-shadow` in Chrome to get equivalent behaviour.

External JavaScript can't show the timer shadow in Firefox, but it works fine on other sites.  Maybe it's a React bug?

wooferzfg:
For some reason, `text-shadow` doesn't seem to work on SVGs in Firefox? Not sure what's going on, but using a `drop-shadow` filter instead of `text-shadow` seems to work in both Firefox and Chrome.

Fix #142

### DIFF
--- a/src/css/Timer.scss
+++ b/src/css/Timer.scss
@@ -5,4 +5,5 @@
   padding-right: 6px;
   font-family: timer, sans-serif;
   margin-top: -10px;
+  text-shadow: none;
 }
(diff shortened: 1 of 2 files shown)


CryZe:
Seems to work well.

PR #187, opened 2019-11-03
Fix regression with split background positioning on Firefox and Edge

### LINKED ISSUE #186: Current Split Marker is positioned incorrectly on Firefox
I guess this happened recently @wooferzfg

wooferzfg:
Fix #186 

We were previously relying on `position: absolute` defaulting to being relative to the table row, but that assumption doesn't hold true on most browsers. I tested this in Chrome, Firefox, and Edge.

### DIFF
--- a/src/css/Splits.scss
+++ b/src/css/Splits.scss
@@ -11,7 +11,6 @@
 
   .split {
     display: table-row;
-    height: 24px;
     line-height: 20px;
 
     @include time;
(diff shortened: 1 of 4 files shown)


CryZe:
Oh yeah, it seems you are multiplying or something. The split index is unrelated to the actual index that we are displaying it at. Especially once we'll have subsplits.

CryZe:
Looks like it's off by one pixel when the column labels are being shown (probably due to the thick separator we use then).

CryZe:
I'm still seeing similar effects when zooming really closely, so it's off by some subpixels, but it's good enough for now.

PR #203, opened 2019-12-01
Fix live validation for segments in the Run Editor

### LINKED ISSUE #8: Live Validation is broken for the Segments
Submitting every valid intermediate state causes the fixing algorithms to run on those intermediate states, which often causes the Best Segment Times and the following segments to get fixed in an unintended way.

### LINKED ISSUE #55: Best Segment Time incorrect when inserting splits via Editor
LiveSplit/livesplit-core#63

wooferzfg:
Changes to times are now submitted in `onBlur` instead of `onChange`.

Fix #8 
Fix #55

### DIFF
--- a/src/ui/RunEditor.tsx
+++ b/src/ui/RunEditor.tsx
@@ -29,10 +29,14 @@ export interface State {
 }
 
 interface RowState {
-    splitTimeIsValid: boolean,
-    segmentTimeIsValid: boolean,
-    bestSegmentTimeIsValid: boolean,
-    comparisonIsValid: boolean[],
+    splitTime: string,
+    splitTimeChanged: boolean,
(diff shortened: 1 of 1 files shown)


src/ui/RunEditor.tsx
```diff
@@ -1510,116 +1510,116 @@ export class RunEditor extends React.Component<Props, State> {
...
+
         this.setState({
             ...this.state,
             editor: this.props.editor.stateAsJson(),
             rowState: {
                 ...this.state.rowState,
-                splitTimeIsValid: true,
-            },
+                splitTimeChanged: false
+            }
```
CryZe: I'd like to keep the trailing commas in general. Just letting you know, it's not something you need to fix.

PR #204, opened 2019-12-18
Add support for custom variables

wooferzfg:
This PR uses the `custom-variables` branch of livesplit-core: https://github.com/LiveSplit/livesplit-core/tree/custom-variables. The branch matches https://github.com/LiveSplit/livesplit-core/pull/201 but has some additional changes that we probably want to carry over into the custom variables livesplit-core PR.

For now, I added the ability to add a custom variable and modify a custom variable. The bindings don't currently have a way to remove a custom variable, so I guess that would have to be added first before I can add removal functionality. I'm also not sure exactly how we'll handle removal because the Settings component doesn't currently support anything like that.

We also want components to be able to display custom variables, but I guess that can come later as well.

### DIFF
--- a/src/livesplit.ts
+++ b/src/livesplit.ts
@@ -515,6 +515,8 @@ export interface KeyValueComponentStateJson {
     key: string,
     /** The value to visualize. */
     value: string,
+    /** The visual color of the delta time. */
+    visual_color: Color,
     /**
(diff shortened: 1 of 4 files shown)


wooferzfg:
Moved over to my own fork with https://github.com/LiveSplit/LiveSplitOne/pull/205

PR #205, opened 2019-12-19
Add support for custom variables

wooferzfg:
This PR uses the changes in https://github.com/LiveSplit/livesplit-core/pull/201

For now, I added the ability to add a custom variable and modify a custom variable. The bindings don't currently have a way to remove a custom variable, so I guess that would have to be added first before I can add removal functionality. I'm also not sure exactly how we'll handle removal because the Settings component doesn't currently support anything like that.

We also want components to be able to display custom variables, but I guess that can come later as well.

### DIFF
--- a/src/css/RunEditor.scss
+++ b/src/css/RunEditor.scss
@@ -242,4 +242,20 @@
       }
     }
   }
+
+  .removable-string-box {
+    display: grid;
+    grid-template-columns: 1fr 30px;
+    column-gap: 5px;
+
+    button {
+      margin: 0px;
+      font-size: 16px;
+      height: 22px;
+    }
+
+    input:focus {
+      outline: 0;
+    }
+  }
 }
--- a/src/layout/Text.tsx
+++ b/src/layout/Text.tsx
@@ -4,6 +4,7 @@ import { gradientToCss, colorToCss } from "../util/ColorUtil";
 import { map } from "../util/OptionUtil";
 
 import "../css/Text.scss";
+import "../css/KeyValue.scss";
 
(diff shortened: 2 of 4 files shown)


CryZe:
> We also want components to be able to display custom variables, but I guess that can come later as well.

Doesn't the text component already support showing them? Pretty sure that's like the first thing I implemented.

wooferzfg:
Oops, you're right. I didn't notice the "Use Variable" option before for some reason.

CryZe:
This seems to delete a bunch of stuff. You probably should've tried rebasing the lsc branch instead of deleting stuff here. I'll try to fix it.

CryZe:
I added removing the variables to the C API.

wooferzfg:
Thanks for doing that + rebasing the branch in livesplit-core. I temporarily removed stuff like the Segment Time component because it would be easy to put back later and I didn't want to deal with the rebase, but now all that's done :grinning: 

I'll work on adding a delete button for each variable next.

wooferzfg:
I added the ability to remove custom variables. This involved a refactor of `Settings.tsx` to use a new interface, `ExtendedSettingsDescriptionJson`, that is an extension of the `SettingsDescriptionJson` defined by `livesplit-core`. This allows us to define additional types of settings that are exclusive to LiveSplit One without needing to modify `livesplit-core`.

CryZe:
Nice, I‘ll take a look when I‘m back from vacation (on the 2nd)

CryZe:
It seems like we don't show any information on the variables tab at all until the speedrun.com information resolved, which means you can't use custom variables for games that are not on speedrun.com (or the name just doesn't match).

PR #210, opened 2020-01-06
Update start:electron npm target

asauber:
The electron app wouldn't run for me without this change.

Is there a build step that I'm missing otherwise?

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "build:core": "node buildCore.js",
     "start": "webpack-dev-server --open",
-    "start:electron": "webpack --electron && electron dist/electron",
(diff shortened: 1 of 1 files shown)


TheEssenceOfDarkness:
>Is there a build step that I'm missing otherwise?

Running this under electron isn't well supported right now in general.

You are welcome to improve this but keep in mind we have very little plans to focus on electron builds right now due to a native desktop app being in the works.

Either way this looks good to me.

CryZe:
Yeah we don't really actively support Electron as we are looking into more native options for the desktop version. If it works like that for you though, I'll merge it, to at least keep it somehow alive.

PR #213, opened 2020-01-15
Simplify delta calculation when scrolling

wooferzfg:


CryZe:
I wonder why we do max / min here anyway, because the Desktop version actually [consumes all scrolls](https://github.com/CryZe/livesplit-one-desktop/blob/498dab9046445303dee4084c09e79fe397b846a7/src/main.rs#L84-L95):
```rust
                let mut scroll = match delta {
                    MouseScrollDelta::LineDelta(_, y) => -y as i32,
                    MouseScrollDelta::PixelDelta(delta) => (delta.y / 15.0) as i32,
                };
                while scroll < 0 {
                    layout.scroll_up();
                    scroll += 1;
                }
                while scroll > 0 {
                    layout.scroll_down();
                    scroll -= 1;
                }
```

wooferzfg:
My mouse only returns values of `-100` and `100` when scrolling the mouse wheel. Is that the same for every mouse / how would we interpret those values?

CryZe:
Probably similar to pixel delta I would assume, except the divisor is 100 or something? We'd have to check what 100 means / the unit of deltaY is.

CryZe:
Looks like it's separated my the mode, just like in the desktop version: https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaMode

CryZe:
Though in both frontends we probably shouldn't discard the "remainder" of the division. Otherwise you may not scroll enough per event and never get any scrolling to happen (on a touchpad / touchscreen for example).

wooferzfg:
So the `100` must be pixels then. It seems kinda weird to me that we would use a hardcoded value like `15` in the desktop version though. I think that different people are able to configure different scroll speeds, and I'm not sure how we could consistently interpret that beyond just getting the scroll direction.

CryZe:
Yeah this gets even more weird because apparently you can set the amount of lines you want to scroll by in the Windows mouse settings. 3 lines seems to be the default, which probably correlates with 5 pixels per line on the Desktop and 33.3... (?) in the browser.

wooferzfg:
Can we merge or close this PR and open a new issue to look into how scrolling works on different mouses?

Issue #129, opened 2018-12-11
The Side Bar doesn't work on iOS anymore.


CryZe:
Gotta love JS libraries.

CryZe:
The feature seems to have been removed: https://github.com/balloob/react-sidebar/issues/67

Their reasoning is that it wasn't really working anyway with the browser also going back a page by swiping, which is true, except for when you pin LSO to the home screen. Well, I guess we wanted to get rid of the sidebar in the long run anyway.

wooferzfg:
Can we close this since you can just open the sidebar by pressing on the button for it now?

CryZe:
Yeah

PR #214, opened 2020-01-15
Always show the sidebar when the screen is wide enough

wooferzfg:


CryZe:
I have yet to try this branch myself. However I think we really should try to replace the sidebar entirely, as it doesn't work at all on iOS anymore #129

wooferzfg:
Could we just have a button open the sidebar instead of relying only on swiping?

CryZe:
Yeah I guess that's okay. It should probably hover in the bottom left corner and only when not in desktop mode or so.

wooferzfg:
I agree that we need to completely rework our layout for mobile. The editors are especially hard to use because you constantly have to scroll. If we implement mobile apps with something like Flutter or React Native, we need to think about how we should maintain both the web and app versions.

For now though, I added a floating button to open the sidebar. This button only shows up when your screen isn't wide enough to show the sidebar by default. In my opinion, it's at least a good temporary improvement over what we currently have.

CryZe:
I just tried this. Is there a reason we are going with it being pinned to the right instead of the left?

CryZe:
Additionally we should modify our margins of the different views to be the same as the layout editor (which apparently has larger margins).

PR #217, opened 2020-01-17
Load React asynchronously

### LINKED ISSUE #216: Delay downloading React
We shouldn't need React in the initial chunk, which really just needs to show the loading screen. By asynchronously importing React, we can start downloading the WASM a bit earlier and thus end the loading screen a bit earlier, as the WASM download is the bottleneck.

wooferzfg:
Fix #216 

Before:
before

After:
after

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,17 +1,18 @@
 const UI = import("./ui/LiveSplit");
+const ReactImport = import("react");
+const ReactDOMImport = import("react-dom");
 const Toastify = import("react-toastify");
-
-import * as React from "react";
-import * as ReactDOM from "react-dom";
 
 import "./css/font-awesome.css";
 import "./css/main.scss";
 
 async function run() {
-    try {
-        const { LiveSplit } = await UI;
-        const { toast, ToastContainer } = await Toastify;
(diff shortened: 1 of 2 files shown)


src/index.tsx
```diff
@@ -34,12 +39,20 @@ Those are known to block WebAssembly.`);
...
+    if (!uiLoaded) {
+        ReactDOM.render(
+            <div className="initial-load">
+                <div className="fa fa-spinner fa-spin"></div>
+                <div className="initial-load-text">Loading...</div>
+            </div>,
+            document.getElementById("base"),
+        );
+    }
+}
```
CryZe: You don't need this function. You can just move this div into the HTML template (src/index.html).

wooferzfg: When I move it into the HTML template, the CSS files that we import in `index.tsx` don't get applied to it, so the loading screen looks like this:

chrome_J3GBhIiNLp

CryZe: You likely kept `className`. You need to rename it to `class`. It's `className` in jsx / tsx because `class` is a keyword.

wooferzfg: oh oops, i'm dumb lol

PR #218, opened 2020-01-18
Hide Run Editor tabs when they are not applicable

### LINKED ISSUE #93: Hide Run Editor Tabs when not applicable
If the speedrun.com database isn't downloaded yet or the game isn't part of the database, the Variables, Rules and the Leaderboard Tab are completely empty and don't need to be shown. If the game is available, but the category is not known, the Variables and Rules Tab at least show some information, while the Leaderboard still doesn't exist. So in that case possibly only the Leaderboard needs to be removed.

Alternatively we could just show some simple message on those tabs, why the information is not available.

wooferzfg:
Fix #93

### DIFF
--- a/src/ui/RunEditor.tsx
+++ b/src/ui/RunEditor.tsx
@@ -115,16 +115,7 @@ export class RunEditor extends React.Component<Props, State> {
 
         const tab = this.getTab();
 
-        let categoryNames = ["Any%", "Low%", "100%"];
-        let category = null;
-        const categoryList = getCategories(this.state.editor.game);
-        if (categoryList != null) {
-            categoryNames = categoryList.map((c) => c.name);
-            const categoryIndex = categoryNames.indexOf(this.state.editor.category);
-            if (categoryIndex >= 0) {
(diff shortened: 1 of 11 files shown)


src/ui/RunEditor.tsx
```diff
@@ -1729,6 +1721,29 @@ export class RunEditor extends React.Component<Props, State> {
...
+    private shouldShowTab(tab: Tab) {
+        if (tab === Tab.RealTime || tab === Tab.GameTime) {
+            return true;
+        }
+
+        const gameInfo = getGameInfo(this.state.editor.game);
+        if (gameInfo == null) {
+            return false;
+        }
+        if (tab === Tab.Variables || tab === Tab.Rules) {
```
CryZe: We may want to show the variables despite there not being any game info (yet). Because otherwise you can't use the custom variables.

CryZe:
Can we show something here instead of nothing at all?



Maybe some explanation that neither custom variables are available nor information about the game is available on speedrun.com.

PR #223, opened 2020-01-19
Set up GitHub Actions

### LINKED ISSUE #174: Switch to GitHub Actions
Travis is having compilation issues now where the gcc installed doesn't support C++ 14. We probably want to switch to GitHub Actions anyway, so this is the issue tracking that.

wooferzfg:
Here's a build that finished which included a deployment step: https://github.com/wooferzfg/livesplitone/commit/b51f1b1987405c1654a2c7945eb30622e6cd8ae6/checks?check_suite_id=408981317

Here's another build that finished which used the cached directories: https://github.com/wooferzfg/livesplitone/commit/b1cf4373ece2560920945c64c0a402548047c6c2/checks?check_suite_id=409000668

Fix #174


.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
...
+    - name: Deploy application
+      if: github.repository == 'LiveSplit/LiveSplitOne' && github.ref == 'refs/heads/master'
+      run: |
+        git config --global user.email "action@github.com"
+        git config --global user.name "GitHub Action"
+        git checkout -b gh-pages
+        mv dist/* .
+        git add :/
+        git commit -m "GitHub Pages build"
+        git push --force https://action:${{ secrets.LIVESPLIT_ONE_DEPLOY }}@github.com/LiveSplit/LiveSplitOne.git
```
CryZe: Is there no prebuilt Action for pushing?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
...
+    - name: Cache target
+      uses: actions/cache@v1
+      with:
+        path: livesplit-core/target
+        key: ${{ runner.os }}-target
+
+    - name: Install Node
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
```
CryZe: I'd prefer not to specify a version at all / always choose the latest Node.

.github/workflows/ci.yml
wooferzfg: There is, but I had trouble getting it to work when I was setting up the build to push the `livesplit-core` docs. I'll try it again though.

.github/workflows/install.sh
```diff
@@ -1,23 +1,15 @@
...
     cd $HOME
     git -C binaryen pull || git clone https://github.com/WebAssembly/binaryen binaryen
     cd binaryen
     cmake .
     make wasm-opt
 
-    rustup target install $TARGET
-
     cargo install -f wasm-gc
     cargo install -f wasm-bindgen-cli
```
CryZe: I believe we may not need the -f anymore. In Rust 1.41 (which this being nightly should be 1.42) it supposedly is automatically updating on cargo install (but only if it's necessary). And since we are caching, this should reduce CI build times a lot.

.github/workflows/ci.yml
wooferzfg: The default version for `actions/setup-node` is Node 10, but I think we want to be on Node 12 (since it's the latest LTS release). I don't think this action lets you specify "latest LTS" any other way besides `12.x`.

.github/workflows/install.sh
CryZe: Seems like this works:

First it upgrades from .57 to .58:


And then it refuses to upgrade any further:

.github/workflows/ci.yml
CryZe: Is there an issue tracking this there? Or should we maybe open one?

wooferzfg: https://github.com/actions/setup-node/issues/26

CryZe: Can you mark it with a comment `FIXME: https://github.com/actions/setup-node/issues/26` so we remember checking this.

CryZe:
The badge in the README should be replaced then too.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - 'master'
+    tags:
+      - '*'
```
CryZe: I don't think we want it on tags.

wooferzfg:
Updated build with deployment: https://github.com/wooferzfg/LiveSplitOne/commit/f6097dd455226c4ab00dd35e2b50e7ba294407a8/checks?check_suite_id=409025713

Updated build with cache: https://github.com/wooferzfg/LiveSplitOne/commit/aec2538f7f327f566ba362f9b0c61b4194f0c947/checks?check_suite_id=409044777

.github/workflows/install.sh
```diff
@@ -1,25 +1,17 @@
...
 
-    rustup target install $TARGET
-
-    cargo install -f wasm-gc
-    cargo install -f wasm-bindgen-cli
+    cargo install wasm-gc
+    cargo install wasm-bindgen-cli
 
     # This fetches latest stable release
     local tag=$(git ls-remote --tags --refs --exit-code https://github.com/japaric/cross \
```
CryZe: Why are we even installing cross? That's from way back when we used emscripten. I believe we can throw this out entirely too.

.github/workflows/install.sh
```diff
@@ -1,25 +1,17 @@
...
-    curl -sSf https://build.travis-ci.org/files/rustup-init.sh | sh -s -- --default-toolchain=nightly -y
-    export PATH=$HOME/.cargo/bin:$PATH
-
-    local target=
-    if [ $TRAVIS_OS_NAME = linux ]; then
-        target=x86_64-unknown-linux-musl
-        sort=sort
-    fi
+    local target=x86_64-unknown-linux-musl
+    local sort=sort
```
CryZe: This stuff up here is also for cross.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,76 @@
...
+        git add :/
+        git commit -m "GitHub Pages build"
+    
+    - name: Deploy application
+      if: github.repository == 'LiveSplit/LiveSplitOne' && github.ref == 'refs/heads/master'
+      uses: ad-m/github-push-action@master
+      with:
+        branch: 'gh-pages'
+        force: true
+        github_token: ${{ secrets.LIVESPLIT_ONE_DEPLOY }}
```
CryZe: The reason why I've been asking if there's not a pre-existing one was because I believe we don't need to provide our own token. GH Actions already come with their own one. So I believe we should use that one and remove this one entirely.

CryZe: https://github.com/LiveSplit/livesplit-core/blob/master/.github/workflows/rust.yml#L677

wooferzfg: Oh yeah, you're totally right. We only needed our own token in `livesplit-core` because we were uploading to a different repo.

Issue #122, opened 2018-10-27
Implement 2 Row Mode for all the Components


CryZe:
livesplit-core has this now, but LSO doesn't render it yet.

PR #224, opened 2020-01-22
Allow KeyValue and Text components to show two rows

wooferzfg:
This fixes part of #122. I'll work on getting two row mode to work for the Splits component later at some point.


src/layout/KeyValue.tsx
```diff
@@ -9,33 +9,53 @@ export interface Props { state: LiveSplit.KeyValueComponentStateJson }
 
 export default class KeyValue extends React.Component<Props> {
     public render() {
+        const keyCell = <td
+            className="key-value-key"
+            style={{
+                color: map(this.props.state.key_color, colorToCss),
```
CryZe: This makes me sad. They added ?. in TypeScript (and JavaScript), which is almost like ? in Rust, except it's completely useless outside of method calls. So this could look like this if `.toCss()` was a method, but it isn't:
```typescript
this.props.state.key_color?.toCss()
```

:(

src/layout/KeyValue.tsx
```diff
@@ -9,33 +9,53 @@ export interface Props { state: LiveSplit.KeyValueComponentStateJson }
...
+                    {valueCell}
+                </tr>
+            </>;
+        } else {
+            keyValueRows = <>
+                <tr>
+                    {keyCell}
+                    {valueCell}
+                </tr>
+            </>;
```
CryZe: I didn't know this is a thing in React. Can we use this in other places as well?

src/css/KeyValue.scss
```diff
@@ -7,8 +7,21 @@
 
   @include time;
 
+  &.key-value-two-rows {
+    height: 40px;
```
CryZe: Can we use scss variables for this or something? For the Desktop version we defined constants for the various common heights, such as the single row height and the two row height: https://github.com/LiveSplit/livesplit-core/blob/master/src/rendering/mod.rs#L108

src/layout/KeyValue.tsx
wooferzfg: I think it's equivalent to an array of elements, so we could use it whenever we do that. However, it actually shouldn't be used in the specific line that you commented on because it's surrounding a single `tr`, and is therefore useless.

src/layout/KeyValue.tsx
wooferzfg: How would you do it in Rust?

TheEssenceOfDarkness: @wooferzfg would be pretty similar to what CryZe described

```rust
//T and U are different types
fn func(&self: T) -> Option<U> {
    self.field?.method()
}
```

wooferzfg:
I refactored the KeyValue and Text components a bit, which also made it easy to add displaying two rows to the Text component.

CryZe:
The Key seems to be moving a bit upwards when switching to 2 rows:

CryZe:
Other than that, it does seem to work well. 👍

wooferzfg:
This is intentional because I reduced the line height when showing two rows (which we also do in the original LiveSplit). I can get rid of that behavior if you think it would be better not to do that though.

CryZe:
We are not doing this in the native renderer in lsc. We just do the same padding on top, then two lines and then the padding at the bottom (0.1 + 0.8 + 0.8 + 0.1 = 1.8 vs. the single row height of 0.1 + 0.8 + 0.1 = 1)

CryZe:
In fact the Title Component is also just considered a two row component, so it follows the exact same positioning rules as the key value component rendering does. So theoretically we should strive those two to be sharing the same CSS long term.

CryZe:
This doesn't seem to work???? I'm not sure what happened, but this is not working on https://one.livesplit.org

wooferzfg:
^ caused by #232

PR #225, opened 2020-01-24
Show variable values as leaderboard columns

### LINKED ISSUE #94: Show Variables as Leaderboard Columns
Just like on speedrun.com

wooferzfg:
Fix #94 

I also made all the run editor tabs be the same width and cleaned up some of the Run Editor logic.

As mentioned on the issue, we maybe want to also show some of the other details as well (date and platform), or potentially jusr show none of this stuff. I'm OK with either route - if we don't want to show variables, then I'll just keep the cleanup stuff in this PR.

### DIFF
--- a/src/ui/RunEditor.tsx
+++ b/src/ui/RunEditor.tsx
@@ -795,7 +795,7 @@ export class RunEditor extends React.Component<Props, State> {
     }
 
     private renderLeaderboard(category: Option<Category>): JSX.Element {
-        const leaderboard = getLeaderboard(this.state.editor.game, this.state.editor.category);
+        const leaderboard = this.getCurrentLeaderboard(category);
         if (leaderboard === undefined) {
             return <div />;
         }
@@ -1312,6 +1312,14 @@ export class RunEditor extends React.Component<Props, State> {
         );
     }
 
+    private getCurrentLeaderboard(category: Option<Category>) {
+        const categoryName = category?.name;
+        if (categoryName) {
+            return getLeaderboard(this.state.editor.game, categoryName);
+        }
+        return undefined;
+    }
+
     private getCurrentCategoriesInfo() {
         let categoryNames = ["Any%", "Low%", "100%"];
         let category = null;
@@ -1745,7 +1753,8 @@ export class RunEditor extends React.Component<Props, State> {
 
         if (tab === Tab.Leaderboard) {
             const { category } = this.getCurrentCategoriesInfo();
-            if (category !== null) {
+            const leaderboard = this.getCurrentLeaderboard(category);
+            if (leaderboard !== undefined) {
                 return true;
             }
         }
--- a/src/ui/RunEditor.tsx
+++ b/src/ui/RunEditor.tsx
@@ -614,7 +614,7 @@ export class RunEditor extends React.Component<Props, State> {
     }
 
     private variableIsValidForCategory(variable: Variable, category: Option<Category>) {
-        return (variable.category == null || variable.category === map(category, (c) => c.id)) &&
+        return (variable.category == null || variable.category === category?.id) &&
             (variable.scope.type === "full-game" || variable.scope.type === "global");
     }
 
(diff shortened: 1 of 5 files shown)


src/ui/RunEditor.tsx
```diff
@@ -616,6 +613,11 @@ export class RunEditor extends React.Component<Props, State> {
         );
     }
 
+    private variableIsValidForCategory(variable: Variable, category: Option<Category>) {
+        return (variable.category == null || variable.category === map(category, (c) => c.id)) &&
```
CryZe: Isn't this just `category?.id` here? though I believe that could yield undefined instead of null, so === may not be what we want. idk, doesn't really matter that much.

src/ui/RunEditor.tsx
```diff
@@ -913,7 +919,7 @@ export class RunEditor extends React.Component<Props, State> {
 
                             expandedRow =
                                 <tr key={`${run.id}_expanded`} className={evenOdd}>
-                                    <td colSpan={4}>
+                                    <td colSpan={4 + (variables?.length || 0)}>
```
CryZe: I believe you can use ?? instead of || here, which is the new way to do it, as it doesn't apply to false-like values.

src/ui/RunEditor.tsx
```diff
@@ -850,33 +855,34 @@ export class RunEditor extends React.Component<Props, State> {
...
                                     if (valueName != null) {
-                                        renderedVariables.push(
+                                        renderedVariablesInExpandedRow.push(
                                             <tr>
                                                 <td>{variable.name}:</td>
                                                 <td>{valueName}</td>
                                             </tr>,
                                         );
-                                    }
+                                    }
```
CryZe: You have some trailing whitespace. You may want to enable trimming trailing whitespace in VS Code.

src/ui/RunEditor.tsx
wooferzfg: Yeah, we should do `category?.id` here. If `variable.category` is null, it'll meet the previous conditional, so we should be ok to just use `===`.

src/ui/RunEditor.tsx
wooferzfg: We should probably add `tslint` to our CI to catch stuff like this.

CryZe: Yeah that would be good.

CryZe:
This doesn't seem to work well when adding new comparisons:

src/ui/RunEditor.tsx
wooferzfg: https://github.com/LiveSplit/LiveSplitOne/pull/226

src/ui/RunEditor.tsx
```diff
@@ -797,7 +792,13 @@ export class RunEditor extends React.Component<Props, State> {
     private renderLeaderboard(category: Option<Category>): JSX.Element {
         const leaderboard = getLeaderboard(this.state.editor.game, this.state.editor.category);
         if (leaderboard === undefined) {
-            return <div />;
+            return <div className="run-editor-tab">
+                {this.getPlaceholderTabContent(
+                    <div className="leaderboard-spinner">
+                        <div className="fa fa-spinner fa-spin" />
+                    </div>,
+                )}
```
CryZe: We can get here? I thought we are not showing the tab anymore when there's no leaderboard?

wooferzfg: We don't show the leaderboard until the game is loaded, but there's still time after that where all the runs need to be loaded. I can change our logic to just hide the tab until that happens rather than showing a spinner.

CryZe: Yeah I'd prefer that.

CryZe: Though a lot of the loading and co. will change once this is happening in lsc I guess.

wooferzfg: We can worry about that when it happens.

CryZe:
This seems really good now. I still found some bug on some obscure game though:

CryZe:
It might be the super long comment that is able to widen the leaderboard.

wooferzfg:
I just pushed some fixes:
- Hide the leaderboard tab when the leaderboard isn't loaded yet rather than showing a spinner.
- Prevent the run comment from expanding the overall leaderboard.
- Change the line height on the markdown (which affects the rules and run comments) to be slightly larger to make it easier to read.
- Fix the bottom of the page not having any padding.

PR #226, opened 2020-01-25
Add tslint to CI

wooferzfg:
We should rebase this branch onto master if it's outdated and get passing CI builds before merging. We should also rebase any open PRs onto master after this is merged and make sure that any linter errors are fixed.


CryZe:
We should maybe use a tslint action or something. This one doesn't seem to create any annotations.

CryZe:
Nice, I manually implemented a matcher and it seems to work:

PR #229, opened 2020-01-25
Key the caches by the current two weeks

### LINKED ISSUE #228: Suboptimal CI Caching
You need to know beforehand if you have a 100% cache hit (which also means to them that you are not going to make any changes to it). If it's not a cache hit it doesn't download anything. So either it downloads and it skips uploading or vice versa. This is pretty pointless and they need to fix it:
https://github.com/actions/cache/issues/106

wooferzfg:
Addresses #228

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,23 +17,26 @@ jobs:
     - name: Checkout submodules
       run: git submodule update --init --recursive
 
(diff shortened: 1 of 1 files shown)


wooferzfg:
oh wait, I forgot that we wanted to make it 2 weeks

wooferzfg:
should be good to go now

PR #230, opened 2020-01-25
Tslint action

wooferzfg:


CryZe:
I'd rather not give random actions the secret token. I think the current solution is fine. Also I'm pretty sure because it is using the secret it is using an API that doesn't even work for Pull Requests. We had the same problem with clippy in lsc.

wooferzfg:
Yeah, this doesn't seem to actually work and every other action out there seems even more messy.

PR #236, opened 2020-01-31
Add buttons to load the default splits and layout

### LINKED ISSUE #234: Loading the Default Layout
So far there's not really a way to load the default layout again. We haven't added this to the sidebar because it already was very cluttered. With the layout submenu, we can nicely introduce this now.

I forgot if we have "default run" in the original LiveSplit, but if we do, we may want that as well.

wooferzfg:
Fix #234

### DIFF
--- a/src/ui/LiveSplit.tsx
+++ b/src/ui/LiveSplit.tsx
@@ -54,11 +54,7 @@ export class LiveSplit extends React.Component<{}, State> {
     constructor(props: {}) {
         super(props);
 
-        const run = Run.new();
-        run.setGameName("Game");
-        run.setCategoryName("Category");
-        run.pushSegment(Segment.new("Time"));
-
+        const run = this.getDefaultRun();
         const timer = expect(
             Timer.new(run),
             "The Default Run should be a valid Run",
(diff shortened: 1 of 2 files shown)


src/ui/LiveSplit.tsx
```diff
@@ -367,6 +359,20 @@ export class LiveSplit extends React.Component<{}, State> {
         exportFile("layout.ls1l", JSON.stringify(layout, null, 4));
     }
 
+    public loadDefaultSplits() {
+        this.setState({
+            ...this.state,
+            timer: this.getTimerForDefaultRun(),
```
CryZe: This creates a whole new timer that the hotkey system doesn't know about, so this breaks the hotkeys. Instead, we have a method on the timer that replaces the underlying run.

src/ui/LiveSplit.tsx
```diff
@@ -367,6 +359,20 @@ export class LiveSplit extends React.Component<{}, State> {
...
+        });
+    }
+
+    public loadDefaultLayout() {
+        this.setState({
+            ...this.state,
+            layout: Layout.defaultLayout(),
```
CryZe: Additionally, you are never disposing any objects. We can't yet integrate with JS's garbage collector. And because of that, the garbage collector can't clean up Rust's native memory. We basically don't get notified that some JS object, such as the old layout or the old run / timer aren't used anymore, so the underlying Rust object and its memory stay alive forever and thus you cause a memory leak. Check the other methods here for how we usually load layouts and runs. We have a bunch of helpers for disposing old objects.

wooferzfg:
I'm trying to also have a `setRun` helper method, but I'm kinda confused about these lines: https://github.com/wooferzfg/LiveSplitOne/blob/load-defaults/src/ui/LiveSplit.tsx#L631-L633

Why do we do `this.state.timer.writeWith` when loading splits from a file but `this.writeWith` when loading splits from Splits.io?

CryZe:
There likely is no reason that we ever use this.state.timer.writeWith when we have a helper method tbh. So feel free to replace those with the calls to the helpers.

Issue #46, opened 2017-10-10
Introduce changing the layout height


LegendofzeldaLF:
So Can You Guys Make It So You Can Zoom Out And In? I Can Barley See Any of My Splits And I Would Love It If We Could Be Able To Zoom Out And In To See More Of Our Splits. Thanks!

emoon:
I'm not sure if this is the case fore you but the way I solve this is to use the web version and simply zoom  in the browser window. While this isn't perhaps ideal it works pretty good.

CryZe:
Actually modifying this is in at least parts blocked by livesplit-core, which has no notion of layout size at the moment. In addition it's not quite clear yet when to just use the browser's size vs. a fixed size chosen by the layout.

CryZe:
I think we should try experimenting with at least horizontally being able to change the layout width (without any livesplit-core support initially, just a setting in LiveSplit One itself).

CryZe:
Alright so you can horizontally stretch the layout now. So I changed the title to tracking the ability to resize vertically as well.

I think this should actually not be too hard. We just need to calculate by how much we need to multiply each "vertical size" such as font sizes and co. by to achieve a certain height.

PR #237, opened 2020-01-31
Allow the layout to be resized horizontally

wooferzfg:
This resolves some of the issues in #46. There's no saving at the moment since this is LSO only and nothing is in livesplit-core.

### DIFF
--- a/src/css/Layout.scss
+++ b/src/css/Layout.scss
@@ -2,10 +2,26 @@
   font-family: "fira", sans-serif;
   font-size: 15px;
   text-shadow: 2px 2px 2px rgba(0, 0, 0, 0.5);
+  position: relative;
   -webkit-touch-callout: none;
   -webkit-user-select: none;
   -khtml-user-select: none;
   -moz-user-select: none;
   -ms-user-select: none;
   user-select: none;
+
+  .resizable-layout {
+    position: absolute;
+    z-index: 50;
+    height: 100% !important;
+
+    .resizable-handle {
+      cursor: e-resize;
+      bottom: -4px;
+      right: -4px;
+      position: absolute;
+      width: 20px;
+      height: 100%;
+    }
+  }
 }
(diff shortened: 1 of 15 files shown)


src/css/Layout.scss
```diff
@@ -1,12 +1,27 @@
...
   font-family: "fira", sans-serif;
   font-size: 15px;
-  width: 300px;
   text-shadow: 2px 2px 2px rgba(0, 0, 0, 0.5);
+  position: relative;
   -webkit-touch-callout: none;
   -webkit-user-select: none;
   -khtml-user-select: none;
   -moz-user-select: none;
   -ms-user-select: none;
```
CryZe: Do we need these? I don't think we support any of these ancient browsers anymore (we should check if any is still needed though)

CryZe:
I feel like we want to adjust the buttons to have the same width:

CryZe:
Other than that, this seems REALLY good. I'm surprised how well this works.

CryZe:
I fixed that. However something that I noticed is that the sidebar (independent of my changes) doesn't work well anymore. If it's over the timer, then the click region of the timer is preferred over the buttons on the sidebar.

CryZe:
I fixed that too. So now I would merge it, but I'd like you to review my changes as well.

wooferzfg:
The button margins were previously based on percentages rather than being consistent, so I fixed that.

src/css/Layout.scss
wooferzfg: Based on the [compatiblity table](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select), most browsers support the `user-select` properly, but a lot of them require it to be `-webkit-user-select`.

chrome_fqQJpyqebA

Based on this, I think we can keep `-webkit-touch-callout` and `-webkit-user-select` but get rid of the rest.

wooferzfg:
@CryZe This looks good to me now. Feel free to merge if you think it's ready.

PR #242, opened 2020-02-02
Update Splits and KeyValue components to use flexbox instead of tables

### LINKED ISSUE #241: Don't use a table for the splits
The splits are currently using a table layout, which means the column values can affect the column boundaries of other splits. This often means that the segment names are cut off when it's not necessary:

Additionally this limits us when trying to implement 2 row mode for the segments. So changing this will benefit #122

wooferzfg:
Fix #241 

We'll still need to do another refactor to get two row mode to work for the Splits, but it should be significantly easier to get to that point from here.

### DIFF
--- a/src/css/KeyValue.scss
+++ b/src/css/KeyValue.scss
@@ -5,6 +5,7 @@
   height: $default-component-height;
   line-height: $default-component-line-height;
   padding: 0px 6px;
+  display: flex;
 
   @include time;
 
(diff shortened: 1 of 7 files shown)


CryZe:
Some weird stuff seems to be happening with the first split here:

CryZe:
Also the value seems to not be vertically aligned properly. Not sure if that changed or always was like this:

wooferzfg:
@CryZe Those issues should be resolved now.
- I added a 1 pixel top padding to split names in the Splits component and keys in the KeyValue component to address the alignment issue.
- We now ignore any empty colums that are at the end of a split's columns array.
- I fixed a regression on Firefox for the width of the control buttons.

PR #244, opened 2020-02-05
Add an option to the Splits to show two rows

### LINKED ISSUE #122: Implement 2 Row Mode for all the Components

wooferzfg:
Fix #122

### DIFF
--- a/src/css/Splits.scss
+++ b/src/css/Splits.scss
@@ -61,15 +61,10 @@ $split-icon-two-row-size: $two-row-height - $split-column-padding;
         object-fit: contain;
         filter: drop-shadow(2px 2px 1px rgba(0, 0, 0, 0.3));
       }
-
-      .split-icon-empty {
-        height: 18px;
-        padding: 0 9px;
-      }
     }
 
     .split-icon-container-empty {
-      padding-right: 0px;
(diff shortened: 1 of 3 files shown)


CryZe:
Other than the icons not increasing their size, this seems perfect.

CryZe:
And I guess the bouncing again:

wooferzfg:
The bouncing is still the same issue as before, so I'm not sure how we can resolve it.

I'll update the icon height to match the full height when displaying two rows.

CryZe:
Can we just use position: absolute for the thicker line or so, such that it doesn't affect the layout of the rest?

CryZe:
@wooferzfg I pushed some fix for the increased icon size. Can you check if it looks good to you?

wooferzfg:
I tried to get rid of the "bouncing" but it makes things look weird and cut off.

Current implementation:

chrome_vJvotkv2kJ

With the "bouncing" fix:

chrome_O7kf0iVaKt

I think we should just keep what we have for now.

wooferzfg:
Besides that, this PR looks good to me now. Feel free to merge if you think it's ready.

Issue #192, opened 2019-11-08
Adjust sizes, dimensions and paddings of the layout based on the native renderer


CryZe:
For the native layout renderer we specified how all the spacing, dimensions, font sizes, paddings and co. should behave across all components. The native renderer works with an abstract component space, but it can be mapped to a default pixel space by multiplying each dimension, size, padding, ... with 24. The CSS for the web version should be adjusted accordingly.

Here's the description of the coordinate spaces: https://github.com/LiveSplit/livesplit-core/blob/f624800a6c3058f2dd47a995751e0d93e605077e/src/rendering/mod.rs#L8-L64

And here are the exact constants used: https://github.com/LiveSplit/livesplit-core/blob/f624800a6c3058f2dd47a995751e0d93e605077e/src/rendering/mod.rs#L103-L118

CryZe:
This is actually not quite that trivial. The web version uses fonts much smaller than the desktop version. Increasing the font sizes causes the left and right texts of key value pairs components and the columns of the splits component to collide with each other much earlier, and in general the web version handles this very badly at the moment (no abbreviations, one split's columns causing other split's names to be cut off early, ...).

CryZe:
We've done a whole lot of CSS fixes since the last comment. Lots of them have contributed to this. So doing one final pass over it that adjusts all the different dimensions and paddings to match the native renderer should be possible now.

PR #250, opened 2020-02-07
Fix styling in Splits and Title

### LINKED ISSUE #246: Fix padding in Splits component in two row mode
The text shouldn't move when you switch into two row mode. I realized that I forgot to include the extra padding like we do here: https://github.com/LiveSplit/LiveSplitOne/blob/e9283dd35a3247483c41a48bdfac0e00f9fdbd16/src/css/KeyValue.scss#L15-L16

### LINKED ISSUE #247: Fix thick separators
They seem to affect the positioning and sizes of the split below at the moment, which causes visible changes as the thick separator appears and disappears when scrolling.

### LINKED ISSUE #248: The title is cut off when there's no attempt count shown

wooferzfg:
The Title component is quite a bit different now, but its sizes should match the KeyValue component.

Fix #246, fix #247, fix #248

### DIFF
--- a/src/css/Splits.scss
+++ b/src/css/Splits.scss
@@ -6,6 +6,8 @@ $split-label-height: 22px;
 $split-column-padding: 6px;
 $split-icon-size: $default-component-height - $split-column-padding;
 $split-icon-two-row-size: $two-row-height - $split-column-padding;
+$thin-separator-width: 1px;
+$thick-separator-width: 2px;
 
 .splits {
   width: 100%;
@@ -17,6 +19,8 @@ $split-icon-two-row-size: $two-row-height - $split-column-padding;
     position: relative;
     width: 100%;
     height: $default-component-height;
+
+    @include time;
 
     &.two-rows {
       height: $two-row-height;
@@ -41,8 +45,6 @@ $split-icon-two-row-size: $two-row-height - $split-column-padding;
         width: $split-icon-two-row-size;
       }
     }
-
-    @include time;
 
     .split-icon-container {
       padding: 2px $split-column-padding;
@@ -139,11 +141,28 @@ $split-icon-two-row-size: $two-row-height - $split-column-padding;
       height: $split-label-height;
     }
 
-    .current-split-background {
+    .current-split-background,
+    .split-borders {
       position: absolute;
       width: 100%;
       height: 100%;
+      top: 0;
       padding: 0;
+      box-sizing: border-box;
+    }
+
+    .split-borders {
+      border-bottom-width: $thin-separator-width;
+      border-top-width: $thin-separator-width;
+      border-bottom-style: solid;
+      border-top-style: solid;
+      top: -$thick-separator-width / 2;
+      height: calc(100% + #{$thick-separator-width} / 2);
+
+      &.separator-above {
+        border-top-width: $thick-separator-width;
+        border-top-style: solid;
+      }
     }
   }
 }
--- a/src/css/Splits.scss
+++ b/src/css/Splits.scss
@@ -24,6 +24,8 @@ $split-icon-two-row-size: $two-row-height - $split-column-padding;
 
       .split-rows {
         flex-wrap: wrap;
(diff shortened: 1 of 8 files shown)


wooferzfg:
Not sure what `game-title-center-fix` was doing before, but it was super buggy, so I'd rather get rid of it:

chrome_MwO3aGJxbk

src/css/Title.scss
```diff
@@ -1,117 +1,104 @@
...
-  }
-
-  div.meta-two-lines > div#lower-row > div.title-attempts {
-    vertical-align: top;
-    padding-bottom: 0;
-    padding-top: 1px;
-  }
+$title-font-size: 13px;
+$attempts-width: 40px;
+$side-padding: 6px;
```
CryZe: This is the same as the column padding and it is used everywhere else as well. We should move this into the variables.scss

(Though this is part of #192 so it doesn't need to be done now)

src/css/Title.scss
```diff
@@ -1,117 +1,104 @@
...
-    padding-top: 1px;
-  }
-
-  div.meta-two-lines > div#lower-row > div.title-attempts {
-    vertical-align: top;
-    padding-bottom: 0;
-    padding-top: 1px;
-  }
+$title-font-size: 13px;
+$attempts-width: 40px;
```
CryZe: Does this need to be a fixed width? Can't we do it like the split times and co. where we dynamically cut off the split names?

CryZe: I guess this may cause problems centering the text then as this value seems to be hardcoded into the centering logic as well.

wooferzfg: This is harder than the split names because we're trying to center the category text relative to the entire component rather than within just the space that isn't part of the attempt count. I'll try to figure something out though.

CryZe:
This doesn't seem quite right:

src/css/Splits.scss
```diff
@@ -40,7 +46,9 @@ $split-icon-two-row-size: $two-row-height - $split-column-padding;
       }
     }
 
-    @include time;
+    &.could-show-separator {
```
CryZe: idk this still feels like a weird hack to me. While it doesn't "flicker" anymore, it still means the last split's icon is smaller than the rest.

wooferzfg: In order to prevent the flickering, we need to compromise in at least one of the cases (either when showing the thick separator or when showing the thin separator) because the amount of size that the split gets is not consistent. In the original LiveSplit, we solve this by making the last split 1 pixel taller and adding a blank pixel row below the thin separator. Let me know if you have a suggestion for a better solution, but there won't be any "clean" solution because it's not possible to maintain both a consistent size and a consistent position when the amount of blank space available is changing.

CryZe: That's why the thick separator is just an "overlay" in the native renderer so it doesn't affect anything.

https://github.com/LiveSplit/livesplit-core/blob/master/src/rendering/component/splits.rs#L156-L169

wooferzfg: I tried doing that and it didn't look good: https://github.com/LiveSplit/LiveSplitOne/pull/244#issuecomment-582683352

CryZe: I'm not sure I see what doesn't look good about it?!

wooferzfg: Maybe it would be better if the border was extended up instead of down though? idk

CryZe: Yes it should be equally growing up as it does down compared to the thin separator it would be

wooferzfg: The icon is getting cut off:

Screenshot_20200207-184010_Chrome

CryZe: Yes then it's probably either positioned too far down or the icons are just too big.

wooferzfg: I'll move the top borders up by a pixel and get rid of the weird padding fix.

CryZe:
Seems good :D

PR #253, opened 2020-02-18
Change styling to mostly match the native renderer

### LINKED ISSUE #192: Adjust sizes, dimensions and paddings of the layout based on the native renderer

wooferzfg:
Fix #192

### DIFF
--- a/src/css/Timer.scss
+++ b/src/css/Timer.scss
@@ -1,11 +1,14 @@
 @import 'variables';
 
+$timer-font-size: 60px;
+$timer-top-margin: -10px;
+
 .timer-time {
-  font-size: 60px;
+  font-size: $timer-font-size;
   text-align: right;
   padding-left: $side-padding;
   padding-right: $side-padding;
   font-family: timer, sans-serif;
-  margin-top: -10px;
+  margin-top: $timer-top-margin;
   text-shadow: none;
 }
--- a/src/css/variables.scss
+++ b/src/css/variables.scss
@@ -2,11 +2,15 @@ $button-margin: 10px;
 $view-margin: 20px;
 
 $default-component-height: 24px;
-$default-component-line-height: 20px;
-$two-row-height: 40px;
-$two-row-line-height: 15px;
-$two-row-extra-padding: 2px;
-$side-padding: 6px;
+$line-height-to-component-height-ratio: 0.8;
+$default-line-height: $default-component-height * $line-height-to-component-height-ratio;
+$font-size-to-line-height-ratio: 0.85;
+$default-font-size: $default-line-height * $font-size-to-line-height-ratio;
+$two-row-height: $default-line-height + $default-component-height;
+$side-padding: $default-component-height * 0.35;
+$vertical-padding: ($default-component-height - $default-line-height) / 2;
+$thick-separator-width: $default-component-height * 0.1;
+$thin-separator-width: $thick-separator-width / 2;
 
 $border-color: #404040;
 
@@ -27,3 +31,9 @@ $light-row-color: #121212;
 
 $selected-row-color: linear-gradient(rgb(51, 115, 244) 0%, rgb(21, 53, 116) 100%);
 $selected-row-hover-color: linear-gradient(hsl(220, 90%, 70%) 0%, hsl(220, 69%, 40%) 100%);
+
+:export {
+  fontSizeToLineHeightRatio: $font-size-to-line-height-ratio;
+  lineHeightToComponentHeightRatio: $line-height-to-component-height-ratio;
+  sidePadding: $side-padding;
+}
--- a/src/layout/DetailedTimer.tsx
+++ b/src/layout/DetailedTimer.tsx
@@ -4,6 +4,11 @@ import { gradientToCss } from "../util/ColorUtil";
 import { renderToSVG } from "./Timer";
 
 import "../css/DetailedTimer.scss";
+import variables from "../css/variables.scss";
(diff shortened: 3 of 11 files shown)


src/css/variables.scss
```diff
@@ -2,11 +2,13 @@ $button-margin: 10px;
...
-$two-row-height: 40px;
-$two-row-line-height: 15px;
-$two-row-extra-padding: 2px;
-$side-padding: 6px;
+$default-line-height: $default-component-height * 0.8;
+$default-font-size: $default-line-height * 0.85;
+$two-row-height: $default-line-height + $default-component-height;
+$side-padding: $default-component-height * 0.35;
+$vertical-padding: ($default-component-height - $default-line-height) / 2;
+$thick-separator-width: $default-line-height * 0.1;
```
CryZe: shouldn't this be default-component-height * 0.1?

src/layout/DetailedTimer.tsx
```diff
@@ -25,6 +25,10 @@ export default class DetailedTimer extends React.Component<Props> {
         }
 
         const totalHeight = this.props.state.timer.height + this.props.state.segment_timer.height;
+        const topRowHeight = totalHeight * 0.55;
+        const bottomRowHeight = totalHeight - topRowHeight;
+        const topRowFontSize = topRowHeight * 0.5 * 0.85;
+        const bottomRowFontSize = bottomRowHeight * 0.5 * 0.8 * 0.85;
```
CryZe: These 0.85 and 0.8 seem a bit magical here. Those are the same as the ones in variables.scss right? I guess they should be named constants in there then.

CryZe: Oh wait, I guess importing from scss to js is probably not possible right? Oof

CryZe:
Looks like the CI script is having problems. The current week of 08 is being interpreted as base 8 due to the prefix 0. But 8 is not a valid digit in base 8.

CryZe:
I fixed the week number bug in the CI on master.

src/layout/DetailedTimer.tsx
wooferzfg: Apparently this works with webpack: https://github.com/LiveSplit/LiveSplitOne/blob/2ad32229541b75187409759379eddf6e1f3f534f/src/css/variables.scss#L35-L39

Pretty sweet 😄

src/css/Timer.scss
```diff
@@ -1,11 +1,14 @@
 @import 'variables';
 
+$timer-height: 60px;
```
CryZe: Is this even used anywhere? You can change the height of the timer in the layout settings, so we probably do a local style anyway and don't use this I would assume.

CryZe: Oh wait, this is for the font-size? Then idk maybe this is used?

CryZe:
Other than possibly what I mentioned and more reviewing, is there more that still needs to be done? I'm not entirely sure why this is still a draft.

wooferzfg:
I marked it as a draft because I would still like to go through all the components one more time to make sure I didn't miss anything, and because I wasn't sure whether we're OK with such a drastic change to the visual appearance (since everything looks bigger now). If you think the current appearance looks good, we can mark it as "ready for review".

CryZe:
> because I wasn't sure whether we're OK with such a drastic change to the visual appearance (since everything looks bigger now)

Yeah I think that's what we want anyway. On mobile all the text looks really small imo and once we have abbreviation support it should be fine.

src/layout/Timer.tsx
```diff
@@ -27,7 +28,7 @@ export function renderToSVG(
         case 4: shiftX = height / 0.71; break;
         default: throw new Error("Unexpected Fraction Length");
     }
-    const x = `${componentWidth - 6 - 0.85 * shiftX}px`;
+    const x = `${componentWidth - parseInt(sidePadding, 10) - 0.85 * shiftX}px`;
```
CryZe: Can we move this parsing out to basically where we import it? Otherwise it just unnecessarily gets run all the time.

src/layout/DetailedTimer.tsx
```diff
@@ -25,6 +30,10 @@ export default class DetailedTimer extends React.Component<Props> {
         }
 
         const totalHeight = this.props.state.timer.height + this.props.state.segment_timer.height;
+        const topRowHeight = totalHeight * 0.55;
+        const bottomRowHeight = totalHeight - topRowHeight;
+        const topRowFontSize = topRowHeight * 0.5 * fontSizeToLineHeightRatio;
```
CryZe: Also why don't we need to parse them here? Are we multiplying strings here?!

wooferzfg: I think it's implicitly converting the string to a number. We should definitely just do the conversion when we import the variable though.

PR #255, opened 2020-02-19
Implement visual regression testing

### LINKED ISSUE #243: Create visual regression tests for the layout renderer
It would be nice to have visual regression testing for rendering the layout, similar to the rendering tests in livesplit-core.

- We'd need a way to render an image of the layout based on layout state data. The easiest approach that I could think of would be to start a server that renders the layout, use Selenium to open a browser to that page, and take a screenshot of it. There's also the concern that everyone uses a different OS and browser, but Docker would be a solution for that. Ideally, we could test Chrome and Firefox (and maybe some other browsers as well? not sure).
- There's lots of good libraries for image comparison. We could use [imghash](https://www.npmjs.com/package/imghash) and [pixelmatch](https://www.npmjs.com/package/pixelmatch).

wooferzfg:
Fix #243


src/ui/LiveSplit.tsx
```diff
@@ -261,31 +268,35 @@ export class LiveSplit extends React.Component<{}, State> {
...
-                        className="sidebar-button fa fa-bars"
-                        onClick={((e: boolean) => this.onSetSidebarOpen(e)) as any}
-                    />
-                }
-                <div
-                    className={viewContainerClassName}
-                    ref={this.containerRef}
+            <>
+                <input id="file-input" type="file" ref={this.fileInputElement} />
+                <a id="file-output" ref={this.fileOutputElement} />
```
CryZe: I get why this is done this way, but I really don't like it :(

wooferzfg: What's the reason that you don't like it?

CryZe: Because the fact that any html elements even exist at all should be an implementation detail of those file utility functions.

wooferzfg: Looks like it's easy enough to hack around it in the test code without having to restructure our actual application code: https://github.com/LiveSplit/LiveSplitOne/blob/3d45d1c6a66438102ec2cf8f518b83f3c607e670/test/rendering-test.js#L48-L56

CryZe:
What more needs to be done here?

wooferzfg:
We just need to rebase and update the hashes.

wooferzfg:
I didn't realize that the side padding is actually a float, so we want to use `parseFloat` instead. Also, none of the styles in `Timer.scss` actually do anything because we use an svg, so I just got rid of that file.

PR #260, opened 2020-02-20
Implement a Service Worker

CryZe:
GitHub Pages usually responds with the responses expiring after 10 minutes, so if you reload LiveSplit One after 10 minutes, you are not getting any cache hits. With a service worker we can permanently store the application data and only when we publish new changes, will the old application cache entries be removed.

This allows full offline support and should make Chrome even offer to install LiveSplit One as a Portable Web App.

### DIFF
--- a/src/ui/LiveSplit.tsx
+++ b/src/ui/LiveSplit.tsx
@@ -61,7 +61,7 @@ export class LiveSplit extends React.Component<{}, State> {
     constructor(props: {}) {
         super(props);
 
-        if ("serviceWorker" in navigator) {
+        if (process.env.NODE_ENV === "production" && "serviceWorker" in navigator) {
             navigator.serviceWorker.register("/service-worker.js");
         }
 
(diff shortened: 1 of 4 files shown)


src/ui/LiveSplit.tsx
```diff
@@ -61,6 +61,10 @@ export class LiveSplit extends React.Component<{}, State> {
     constructor(props: {}) {
         super(props);
 
+        if ("serviceWorker" in navigator) {
+            navigator.serviceWorker.register("/service-worker.js");
+        }
+
```
CryZe: This is slightly suboptimal as we don't install the service worker in a non-production build, so this 404s in a local environment, as it should. But it still looks bad in the debugger.

CryZe:
This may or may not resolve #260

src/ui/LiveSplit.tsx
wooferzfg: We can add `process.env.NODE_ENV === 'production'` to the `if` condition.

wooferzfg:
This looks good to me besides the 404 issue in a development environment.

PR #266, opened 2020-02-22
Remove the Roboto font

### LINKED ISSUE #265: Roboto font isn't cached by the Service Worker
It's hosted on Google. Either we explicitly cache it, or just host it ourselves. Keep in mind that the woff2 at the bottom of the screenshot also needs to be cached.

wooferzfg:
Fix #265

### DIFF
--- a/src/css/Table.scss
+++ b/src/css/Table.scss
@@ -46,7 +46,7 @@
   .number {
     text-align: right;
     font-weight: bold;
-    font-family: "Roboto", sans-serif;
+    font-variant-numeric: tabular-nums;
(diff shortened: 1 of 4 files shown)


CryZe:
I'm suprised only one test is failing tbh.

CryZe:
Also we should make it so the test failures get annotated in the code, like in lsc.

wooferzfg:
I think the reason that only one test is failing is because image hashing just doesn't work that well. There seem to be a lot of false positives/negatives with it for some reason.

PR #268, opened 2020-02-22
Improve regression testing

wooferzfg:
- Use longer hashes so we have fewer false passing tests.
- Add support for annotations.
- Treat test failures as warnings instead of errors since they're somewhat unreliable.
- Run tests after deployment so that we don't slow down the CD process.


wooferzfg:
Warnings seem to be working: https://github.com/LiveSplit/LiveSplitOne/actions/runs/43559024

PR #269, opened 2020-02-23
Improve UI consistency and mobile layout

### LINKED ISSUE #38: Improve Mobile Layout
As long as we don't have mobile apps, the web version is the way to go, so we should improve the Mobile Layout. 

- [x] Layout Renderer
- [ ] Run Editor
- [ ] Layout Editor
- [ ] Settings

wooferzfg:
Probably the coolest part of these changes is this mixin for mobile logic:
```
@mixin mobile {
  .is-mobile & {
    @content;
  }
}
```
This allows us to implement mobile logic throughout the SCSS rather than restricting it to a separate file. This makes conditional mobile logic super easy and clean compared to what it would be otherwise.

### Tasks:
- [X] Update UI margins to be consistent across all of our editors.
- [X] Update Settings Editor to be mobile friendly
- [x] Update Layout Editor to be mobile friendly
- [x] Update Splits Editor to be mobile friendly

Fix #38

### DIFF
--- a/src/css/ContextMenu.scss
+++ b/src/css/ContextMenu.scss
@@ -3,7 +3,7 @@
 @mixin context-menu {
   .react-contextmenu {
     min-width: 160px;
-    padding: 5px 0;
+    padding: $ui-margin/2 0;
     margin: 2px 0 0;
     font-size: 16px;
     text-align: left;
@@ -23,7 +23,7 @@
     }
 
     .react-contextmenu-item {
-      padding: 3px 10px;
+      padding: $ui-margin/2 $ui-margin;
       font-weight: 400;
       text-align: inherit;
       white-space: nowrap;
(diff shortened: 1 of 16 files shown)


CryZe:
Looks pretty good at first glance. The fonts seem a bit small (at least on my PC, I didn't check on my phone yet). I'll do a more thorough review tomorrow.

CryZe:
Also the layout editor settings don't seem to be full width:

CryZe:
Looks like the overall padding shrank too. Also there's some other changes, not sure if those were intended either.

wooferzfg:
The changes in that last GIF were intentional (and are a part of the "update UI margins to be consistent across all of our editors" task). My goal was to make the margin be 10 pixels everywhere throughout the UI.

We can revert some of the changes if you think it looks bad, but I personally prefer the more consistent look.

CryZe:
Yes the main content should have more padding around it (I even believe the one we had before was maybe not enough). And the sidebar buttons are definitely not consistent. If you look above the sidebar separators, there‘s more spacing there to the next button that when the separator is above a button.

wooferzfg:
I fixed the sidebar separators and I made the margin around the main content be 30 pixels (only on desktop though, because I don't want to limit the available space on mobile).

wooferzfg:
I added larger margins to separate major parts of the UI in each editor. I decided to use these in the mobile view as well, but only to separate the different sections vertically (which helps with readability without taking much of the necessary screen space).

PR #271, opened 2020-02-23
Fix size of game icon to match the two row split icon size

### LINKED ISSUE #270: Title and 2 Row Split Text is unaligned
If both the title and the splits show an icon and the splits are in 2 row mode, their text should be aligned to each other.

wooferzfg:
Fix #270


CryZe:
Shouldn't this possibly fail on some UI test?

wooferzfg:
Oh yeah, it did. We need to update those hashes.

PR #273, opened 2020-02-23
Implement Support for Abbreviations

CryZe:
This implements abbreviations. For the longest time we thought it's not possible to implement them in the web, but it turns out that it can work.

This implements support for the key value pair components so far. Abbreviations are still missing for the Title component. This will need some additional support from livesplit-core.

### DIFF
--- a/src/css/Title.scss
+++ b/src/css/Title.scss
@@ -40,7 +40,7 @@ $game-icon-size: $two-row-height - $vertical-padding * 2;
       height: 50%;
       padding: 0 $side-padding;
 
-      &.centered div {
+      &.centered .abbreviated {
         text-align: center;
       }
     }
@@ -66,7 +66,7 @@ $game-icon-size: $two-row-height - $vertical-padding * 2;
         overflow: hidden;
         padding: 0 $side-padding;
 
-        &.centered div {
+        &.centered .abbreviated {
           text-align: center;
         }
       }
--- a/src/css/Title.scss
+++ b/src/css/Title.scss
@@ -40,8 +40,8 @@ $game-icon-size: $two-row-height - $vertical-padding * 2;
       height: 50%;
       padding: 0 $side-padding;
 
-      &.centered {
-        justify-content: center;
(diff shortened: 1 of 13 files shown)


CryZe:
The current ResizeObserver polyfill here seems to use requestAnimationFrame to query the element's dimensions every frame. This is a reasonable polyfill for platforms that don't have ResizeObserver, but I'm surprised it still does this on my Chrome, which should have proper support for ResizeObserver, so maybe we want something that only falls back to requestAnimationFrame for browsers that don't have support for the proper one.

CryZe:
Another problem here is that clientWidth and scrollWidth are integers, not floating point numbers. So there's a small chance that it still introduces ellipsis even if it could use an abbreviation. It seems fairly rare though, so it's probably fine.

CryZe:
This is blocked by the PR in livesplit-core: https://github.com/LiveSplit/livesplit-core/pull/307

But other than that, it's already reviewable.

src/css/Title.scss
```diff
@@ -40,8 +40,8 @@ $game-icon-size: $two-row-height - $vertical-padding * 2;
       height: 50%;
       padding: 0 $side-padding;
 
-      &.centered {
-        justify-content: center;
+      &.centered div {
```
wooferzfg: Can we do `&.centered .abbreviated` here instead?

src/css/Title.scss
```diff
@@ -66,8 +66,8 @@ $game-icon-size: $two-row-height - $vertical-padding * 2;
         overflow: hidden;
         padding: 0 $side-padding;
 
-        &.centered {
-          justify-content: center;
+        &.centered div {
```
wooferzfg: same comment here

CryZe:
@wooferzfg Can you check what's wrong with the tests? They seem to be timing out.

CryZe:
Feel free to just push the fix and merge this then if you don't see any more problems with this PR.

PR #274, opened 2020-02-26
Create an About dialog

### LINKED ISSUE #28: Add an About dialog
A simple About dialog to show copyright, creators and a link to the GitHub repo, similar to LiveSplit's.

wooferzfg:
In terms of branding, I think we should stick to "LiveSplit One" everywhere rather than doing something like "LiveSplit One Web" or "LiveSplit One Desktop" (although we should still use these internally to name our repos). This kind of goes along with the whole "One" theme - LiveSplit One is a single application that works on a lot of different platforms, and the different contexts (web, desktop) are just different ways to use it.

Fix #28

### DIFF
--- a/src/ui/About.tsx
+++ b/src/ui/About.tsx
@@ -1,0 +1,42 @@
+import * as React from "react";
+
+import LiveSplitIcon from "../assets/icon_small.png";
+
+import "../css/About.scss";
+
+export class About extends React.Component {
+    public render() {
+        return (
+            <div className="about">
+                <div className="about-inner-container">
+                    <div className="livesplit-title">
+                        <span className="livesplit-icon">
+                            <img src={LiveSplitIcon} alt="LiveSplit Logo" />
+                        </span>
+                        <div className="title-text">LiveSplit One</div>
+                    </div>
+                    <p className="build-version">
+                        <a href={`https://github.com/LiveSplit/LiveSplitOne/commit/${COMMIT_HASH}`} target="_blank">
+                            Version: {COMMIT_HASH} ({BUILD_DATE})
+                        </a>
+                    </p>
+                    <p>LiveSplit One is a multiplatform version of LiveSplit, the sleek,
+                    highly-customizable timer for speedrunners.</p>
+                    <p>
+                        <a href="https://github.com/LiveSplit/LiveSplitOne" target="_blank">
+                            View Source Code on GitHub
+                        </a>
+                    </p>
+                    <h1 className="contributors-header">Contributors</h1>
+                    {
+                        CONTRIBUTORS_LIST.map((contributor) => (
+                            <p className="contributor">
+                                <a href={`https://github.com/${contributor}`} target="_blank">{contributor}</a>
+                            </p>
+                        ))
+                    }
+                </div>
+            </div>
+        );
+    }
+}
--- a/src/ui/LiveSplit.tsx
+++ b/src/ui/LiveSplit.tsx
@@ -13,6 +13,7 @@ import * as SplitsIO from "../util/SplitsIO";
 import { LayoutEditor as LayoutEditorComponent } from "./LayoutEditor";
 import { RunEditor as RunEditorComponent } from "./RunEditor";
 import { SettingsEditor as SettingsEditorComponent } from "./SettingsEditor";
+import { About } from "./About";
 import { SideBarContent } from "./SideBarContent";
 import { toast } from "react-toastify";
 
@@ -26,6 +27,7 @@ export enum MenuKind {
     Layout,
     LayoutEditor,
     SettingsEditor,
+    About,
 }
 
 type Menu =
@@ -34,7 +36,8 @@ type Menu =
     { kind: MenuKind.RunEditor, editor: RunEditor } |
     { kind: MenuKind.Layout } |
     { kind: MenuKind.LayoutEditor, editor: LayoutEditor } |
-    { kind: MenuKind.SettingsEditor, config: HotkeyConfig };
+    { kind: MenuKind.SettingsEditor, config: HotkeyConfig } |
+    { kind: MenuKind.About };
 
 export interface State {
     hotkeySystem: HotkeySystem,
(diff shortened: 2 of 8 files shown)


CryZe:
Yeah branding wise it's just LiveSplit One. Sometimes when referring to a specific version though, you still need to call it "the web version of LiveSplit One" or so, as otherwise it's just too ambiguous. Also atm I'd like to point out that it is the web version in order to highlight that the idea is that it's not the only version.

src/ui/About.tsx
```diff
@@ -0,0 +1,34 @@
...
+    public render() {
+        return (
+            <div className="about">
+                <div className="livesplit-title">
+                    <span className="livesplit-icon">
+                        <img src={LiveSplitIcon} alt="LiveSplit Logo" />
+                    </span>
+                    <div className="title-text">LiveSplit One</div>
+                </div>
+                <p className="build-version">Version: {COMMIT_HASH} ({BUILD_DATE})</p>
```
CryZe: I feel like we should make this a clickable link that links to the specific commit.

src/ui/About.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from "react";
+
+import LiveSplitIcon from "../assets/icon_small.png";
```
CryZe: Is this icon inlined into the HTML or is it dynamically loaded at runtime? If it is dynamically loaded, then we may need to consider how this affects the PWA. The assets folder is explicitly cached at runtime. So if you never visited the about page while online, it fails to load this icon offline if it's not inlined.

webpack.config.js
```diff
@@ -1,10 +1,41 @@
...
+        const contributorsData = await fetch(`https://api.github.com/repos/LiveSplit/${repoName}/contributors`);
+        const contributorsJson = await contributorsData.json();
+        return contributorsJson.filter((user) => user.type === "User");
+    }
+
+    const lsoContributors = await getContributorsForRepo("LiveSplitOne");
+    const coreContributors = await getContributorsForRepo("livesplit-core");
+    const allContributors = lsoContributors.reduce(
+        (accumulator, currentUser) => {
+            const existingUser = accumulator.find((user) => user.login === currentUser.login);
```
CryZe: That almost makes me feel like we should be using a map instead of an array here. This is O(n²), though it probably doesn't matter much.

CryZe: Also I feel like reduce is super bloaty, this would likely be shorter (and faster) as a normal loop.

CryZe:
Also idk, I think it looks "really bad" on mobile. Seems like it's a word document or something, not much of a website / UI.

src/ui/About.tsx
```diff
@@ -0,0 +1,34 @@
...
+                    <span className="livesplit-icon">
+                        <img src={LiveSplitIcon} alt="LiveSplit Logo" />
+                    </span>
+                    <div className="title-text">LiveSplit One</div>
+                </div>
+                <p className="build-version">Version: {COMMIT_HASH} ({BUILD_DATE})</p>
+                <p>LiveSplit One is a multiplatform version of LiveSplit, the sleek,
+                    highly-customizable timer for speedrunners.</p>
+                <p>
+                    <a href="https://github.com/LiveSplit/LiveSplitOne">View Source Code on GitHub</a>
```
CryZe: I think these links should open in a new tab.

src/ui/About.tsx
```diff
@@ -0,0 +1,34 @@
...
+                </p>
+                <h1 className="contributors-header">Contributors</h1>
+                {
+                    CONTRIBUTORS_LIST.map((contributor) => (
+                        <p className="contributor">
+                            <a href={`https://github.com/${contributor}`}>{contributor}</a>
```
CryZe: Same here.

webpack.config.js
```diff
@@ -1,10 +1,41 @@
...
+            return accumulator;
+        },
+        coreContributors,
+    );
+
+    const contributorsList = allContributors
+        .sort((user1, user2) => user2.contributions - user1.contributions)
+        .map((user) => user.login);
+    const commitHash = execSync("git rev-parse --short HEAD").toString();
+    const date = moment.utc().format("YYYY-M-D k:mm:ss");
```
CryZe: Month and day should probably be zero padded.

src/ui/About.tsx
CryZe: It is indeed inlined. Though do we really need two icons? That will increase the download size / offline storage.

CryZe: I believe we even want the icon to be smaller anyway. The about icon is much bigger than the text next to it, which looks kind of weird.

wooferzfg: I'm confused what you mean by "two icons". The larger icon is used for the favicon, so I don't think it's ever actually loaded directly. I don't think we should be using the larger icon for other stuff because then it'll get loaded in its full size.

webpack.config.js
wooferzfg: I changed my approach to use loops instead. Let me know if you still see ways to improve it.

src/ui/LiveSplit.tsx
```diff
@@ -295,10 +288,15 @@ export class LiveSplit extends React.Component<{}, State> {
...
-    public openTimerView() {
+    public openTimerView(remount: boolean) {
         this.setState({
             menu: { kind: MenuKind.Timer },
+            sidebarOpen: false,
         });
+        if (remount) {
+            this.state.layout.remount();
+            this.state.hotkeySystem.activate();
+        }
```
CryZe: Do we know why the others don't need it? This seems like a hacky workaround this way. They all need to remount the layout. Or at least they should. I don't see why the Run Editor and the About view would behave differently here. Neither is showing the layout.

webpack.config.js
CryZe: Seems good this way.

src/ui/About.tsx
CryZe: Uhm yeah idk what I was seeing. It looked like this PR was adding yet another icon when I previously reviewed it.

src/ui/LiveSplit.tsx
CryZe: Oh wait, this is just so the submenus don't remount right? I guess we do the run editor and layout editor remounting somewhere else then?!

wooferzfg: The run editor and layout editor have their own methods (`closeRunEditor` and `closeLayoutEditor`). `openTimerView` is only used when you press the "Back" button in the sidebar in the Splits, Layout, and About views.

CryZe: It's still questionable that out of all three only the about view is doing the remounting this way. I feel like we should unify them better.

CryZe:
I'm getting a horizontal scroll bar in the mobile view.

Issue #70, opened 2018-02-22
The 5MB localStorage quota may cause issues with saving splits.


DarkRTA:
As a lot of you may already know, `localStorage` normally has a quota of 5MB, (can vary across browsers), and attempting to store more than that limit will result in an error. In Firefox and Opera, this limit can be raised as needed by the user, but other browsers have to deal with this limit. Currently LiveSplit One is saving and loading splits from `localStorage`, and split files like to grow when icons are added and attempts are recorded. This will cause issues when saving, meaning users will eventually have to start importing and exporting their splits in order to continue using them.

TheEssenceOfDarkness:
There are a few ways to remedy this problem such as removing icons from the splits, or clearing the run history.

wooferzfg:
I think we want to shrink icons (in both LS and LSO) if they're above a certain size when setting them in the Splits Editor or loading splits.

TheEssenceOfDarkness:
I say shrinking them to 64x64 is a good balance between icon quality and file size, that way they still look nice when displayed on the detailed timer.

hallcristobal:
Shrinking them is ideal, compressing and then shrinking less is also good. 
If I remember correctly CryZe went to implement image resizing and it made the final library massive, so a way to do it without as much overhead would be the best.

hallcristobal:
LiveSplit/livesplit-core#47

CryZe:
We now do image shrinking, so this is less of an issue now. We could look into compressing the splits file in local storage, but I think this is mostly a non-issue anymore until we potentially store multiple splits files via the Run Selection View #189

CryZe:
We can also use the native file system api to get raw access to the user's file system:
https://web.dev/native-file-system/

CryZe:
Seems like IndexedDB allows storing MUCH more data (hundreds of megabytes even) without any problem (or even asking?! why does the browser not limit this?). So switching to IndexedDB should resolve this issue entirely.

PR #282, opened 2020-03-06
Switch to IndexedDB

### LINKED ISSUE #256: Switch to IndexedDB
IndexedDB allows storing raw objects that don't need to be serialized to a string. So we could store the splits as raw byte buffers instead, which we need in livesplit-core anyway, rather than JS strings. 
Additionally it seems like browsers allow it to be larger than normal localStorage, so this would also help with #70 

It seems like the API for IndexedDB is quite obscure though, so even the documentation for it recommends using something like [localForage](https://localforage.github.io/localForage/) instead which wraps it nicely. Though it brings a lot of baggage of trying to work on older browsers, but we basically only target browsers that at least support WebAssembly. So those definitely support IndexedDB afaict, so shipping any kind of polyfills is just bloating our download size.

wooferzfg:
Fix #256

I'm opening this as a draft PR because I'm not sure how to approach storing the data as raw bytes or implementing compression for the splits files.

### DIFF
--- a/src/storage/index.tsx
+++ b/src/storage/index.tsx
@@ -12,15 +12,31 @@ async function getDb(): Promise<IDBPDatabase<unknown>> {
     if (db == null) {
         db = await openDB("LiveSplit", 1, {
             upgrade(db) {
-                db.createObjectStore("splits", {
+                const store = db.createObjectStore("settings", {
                     autoIncrement: true,
                 });
-                db.createObjectStore("layouts", {
-                    autoIncrement: true,
-                });
-                db.createObjectStore("settings", {
-                    autoIncrement: true,
-                });
+
+                const splits = localStorage.getItem("splits");
+                if (splits) {
+                    store.put(new TextEncoder().encode(splits), "splits");
+                }
+
+                const layout = localStorage.getItem("layout");
+                if (layout) {
+                    store.put(JSON.parse(layout), "layout");
+                }
+
+                const hotkeys = localStorage.getItem("settings");
+                if (hotkeys) {
+                    store.put(JSON.parse(hotkeys).hotkeys, "hotkeys");
+                }
+
+                const layoutWidth = localStorage.getItem("layoutWidth");
+                if (layoutWidth) {
+                    store.put(+layoutWidth, "layoutWidth");
+                }
+
+                localStorage.clear();
             },
         });
     }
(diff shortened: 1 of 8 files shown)


DarkRTA:
This will also close #70 as it'll no longer apply after merging this.

src/ui/LiveSplit.tsx
```diff
@@ -39,6 +40,13 @@ type Menu =
     { kind: MenuKind.SettingsEditor, config: HotkeyConfig } |
     { kind: MenuKind.About };
 
+export interface Props {
+    splits: string,
+    layout: any,
+    settings: any,
```
CryZe: These probably shouldn't be any.

src/ui/LiveSplit.tsx
```diff
@@ -39,6 +40,13 @@ type Menu =
     { kind: MenuKind.SettingsEditor, config: HotkeyConfig } |
     { kind: MenuKind.About };
 
+export interface Props {
+    splits: string,
```
CryZe: The splits should be ArrayBuffer / Uint8Array.

src/ui/LiveSplit.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from "react";
+import localforage from "localforage";
```
CryZe: I'm not sure if we want localforage or not. If we manually talk to IndexedDB we could have tables for splits and for layouts, which would help us for the run and layout selection view.

CryZe:
Also considering I'm seeing a lot of people use LSO by now, I believe we need to have backwards compatibility with localStorage, so people don't lose their splits.

src/ui/LiveSplit.tsx
CryZe: In fact I believe that we even should "ban any". They somewhat recently introduced `unknown` as a [top type](https://en.wikipedia.org/wiki/Top_type) that doesn't imply dynamic typing.

CryZe:
I just implemented the byte buffer storing.

wooferzfg:
We still need to implement backwards compatibility for localStorage so that splits, layouts, and settings aren't lost once this is merged.

wooferzfg:
Let's make sure that the livesplit-core commits are on master before merging this.

src/storage/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+                db.createObjectStore("splits", {
+                    autoIncrement: true,
+                });
+                db.createObjectStore("layouts", {
+                    autoIncrement: true,
+                });
+                db.createObjectStore("settings", {
+                    autoIncrement: true,
+                });
+            },
```
CryZe: I probably should've told you, but IndexedDB has a notion of migrating between different versions. The migration from no IndexedDB to version 1 is also a migration. This is being done here. And this is the only place we ever need to migrate from localStorage to IndexedDB. So all the migration code should be in here.

wooferzfg: So would we just load everything from localStorage inside `upgrade` and immediately add them to the database?

CryZe: Yes, I'll push a fix for it in a few minutes.

PR #283, opened 2020-03-06
Fix sidebar spacing

wooferzfg:
This fixes the sidebar buttons moving around as you go between some of the views.

Before:

qDYtahoHqD

After:

9ZKPAIFHvZ

### DIFF
--- a/src/ui/SideBarContent.tsx
+++ b/src/ui/SideBarContent.tsx
@@ -65,7 +65,7 @@ export class SideBarContent extends React.Component<Props, State> {
             case MenuKind.Splits: {
                 sidebarContent = (
                     <div className="sidebar-buttons">
-                        <h2>Splits</h2>
+                        <h1>Splits</h1>
                         <hr />
                         <button onClick={(_) => this.props.callbacks.openRunEditor()}>
                             <i className="fa fa-edit" aria-hidden="true" /> Edit
@@ -99,7 +99,7 @@ export class SideBarContent extends React.Component<Props, State> {
             case MenuKind.RunEditor: {
                 sidebarContent = (
                     <div className="sidebar-buttons">
-                        <h2>Splits Editor</h2>
+                        <h1>Splits Editor</h1>
                         <hr />
                         <div className="small">
                             <button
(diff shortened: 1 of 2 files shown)


src/ui/SideBarContent.tsx
```diff
@@ -214,15 +214,15 @@ export class SideBarContent extends React.Component<Props, State> {
...
+                        <hr className="livesplit-title-separator" />
                         <button onClick={(_) => this.props.callbacks.openSplitsView()}>
                             <i className="fa fa-list" aria-hidden="true" /> Splits
                         </button>
                         <button onClick={(_) => this.props.callbacks.openLayoutView()}>
                             <i className="fa fa-layer-group" aria-hidden="true" /> Layout
                         </button>
                         <hr />
-                        <h2>Compare Against</h2>
+                        <h2 className="subheader">Compare Against</h2>
```
CryZe: Can we use h2 here and h1 for the actual menu title? (or h2 / h3)

wooferzfg: I don't think we want to make the Compare Against text any smaller because the text on the buttons is pretty large already. We can probably make the title text bigger though.

CryZe: I'm not talking about changing any sizes. Just using the appropriate header tags instead of a class.

wooferzfg: The header tags come with some default styling though (`h1` actually has a bigger font size but smaller margins than `h2` by default), so we shouldn't use different header tags if we want the font sizes to be the same.

wooferzfg: We could use custom styling to override the margins and font size for `h1` and `h2` so that they're the same. But that seems a bit weird, and I feel like we should just use classes and `span`s in that case.

CryZe: idk, I'd prefer just modifying h1 and h2 to be the way we want them to behave instead of using classes. It probably doesn't matter much, but I think it just semantically makes more sense. Otherwise we might as well just use div for everything.

PR #284, opened 2020-03-08
Implement Gamepad Support

CryZe:
This implements gamepad support for the hotkeys. This allows controlling LiveSplit One without having to focus the browser tab. There seem to be some unfortunate limitations though. Chrome for example acquires exclusive access to the gamepads (at least on Windows with DirectInput). Firefox however doesn't, so you likely would want to use Firefox if Chrome is causing problems with your controller.

The PR in livesplit-core needs to be merged first: https://github.com/LiveSplit/livesplit-core/pull/310

### DIFF
--- a/src/ui/HotkeyButton.tsx
+++ b/src/ui/HotkeyButton.tsx
@@ -41,21 +41,22 @@ export class HotkeyButton extends React.Component<Props, State> {
 
                             let gamepadIdx = 0;
                             for (const gamepad of gamepads) {
-                                if (gamepad === null) { continue; }
+                                if (gamepadIdx >= oldButtonState.length) {
+                                    oldButtonState[gamepadIdx] = [];
+                                }
 
-                                let buttonIdx = 0;
-                                for (const button of gamepad.buttons) {
-                                    const oldState = oldButtonState[gamepadIdx]?.[buttonIdx] ?? false;
-                                    if (button.pressed && !oldState) {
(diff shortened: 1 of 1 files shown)


src/ui/HotkeyButton.tsx
```diff
@@ -33,20 +35,49 @@ export class HotkeyButton extends React.Component<Props, State> {
...
+                                if (gamepad === null) { continue; }
+
+                                let buttonIdx = 0;
+                                for (const button of gamepad.buttons) {
+                                    const oldState = oldButtonState[gamepadIdx]?.[buttonIdx] ?? false;
+                                    if (button.pressed && !oldState) {
+                                        this.props.setValue(`Gamepad${buttonIdx}`);
+                                    }
+
+                                    if (oldButtonState[gamepadIdx] === undefined) {
```
CryZe: > Expression is always false.

No it's not, lmao. I love false positives on linters.

CryZe: Though this is kind of weird, because I feel like not even TypeScript itself models this properly. If you index out of bounds, there's no exception in JS and instead it's just `undefined`. However TypeScript doesn't understand that it seems. Probably for convenience reasons.

wooferzfg:
Works well with my DS3 controller (which uses MotioninJoy to emulate XInput) on both Chrome and Firefox.

PR #291, opened 2020-03-16
Fix linter running in CI without building livesplit-core

wooferzfg:
We committed a linter error to master that wasn't caught by the CI because we were previously running `npm run lint` without building livesplit-core. This made the command innacurate to a proper development environment.

Moving the linter to the normal build ensures that the full setup can happen before we run the linter. This also reduces the amount of setup repetition in the `ci.yml` file. One downside (maybe) is that the linter becomes a blocker for deployment.

[Here's a build that catches the linter error.](https://github.com/LiveSplit/LiveSplitOne/runs/510162370)

@CryZe Let me know your thoughts on what approach we should take for this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,12 @@ jobs:
         set -e
         npm run build:core
 
+    - name: Set up tslint matcher
+      run: echo "::add-matcher::.github/workflows/tslint.json"
+
+    - name: Run tslint
+      run: npm run lint
+
     - name: Build Frontend
       run: |
         set -e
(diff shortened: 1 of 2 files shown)


src/ui/LiveSplit.tsx
```diff
@@ -100,7 +100,7 @@ export class LiveSplit extends React.Component<Props, State> {
                 }
             }
         } catch (_) { /* Looks like the storage has no valid data */ }
-        if (hotkeySystem == null) {
+        if (hotkeySystem === null) {
```
CryZe: I don't want this. `== null` / `!= null` has its place (for checking against both null and undefined) and in fact my local tslint is totally fine with it (unlike == in other cases). So can we please change the linter settings here to allow it in those cases? (I haven't checked if this particular case even could have both null and undefined. If it doesn't, then I'm possibly fine with the change)

PR #292, opened 2020-03-17
Implement a Splits Selection View

CryZe:
Here's a draft PR for what will be the splits selection view. There's some initial changes to the IndexedDB and already shows some very unstyled splits selection view.

- [x] Changes to the IndexedDB
- [x] Basic splits selection
- [x] Basic styling
- [x] Splits Editor for runs that are not loaded
- [x] Modify the open splits key when switching between splits
- [x] Have a notion of an undefined / null splits key for splits that are open, but have never been saved to the DB.
- [x] Store the splits key in the IndexedDB settings table.
~~Handle all the potential race conditions like downloading splits and then closing the splits selection view while the download is happening.~~

### DIFF
--- a/src/api/GameList.ts
+++ b/src/api/GameList.ts
@@ -20,7 +20,7 @@ const gameIdToCategoriesPromises: Map<string, Promise<Category[]>> = new Map();
 const gameIdToGameInfoMap: Map<string, Game> = new Map();
 const gameIdToGameInfoPromises: Map<string, Promise<Game>> = new Map();
 const gameAndCategoryToLeaderboardPromises: Map<string, Promise<void>> = new Map();
-const gameAndCategoryToLeaderboardMap: Map<string, Run<PlayersEmbedded>[]> = new Map();
+const gameAndCategoryToLeaderboardMap: Map<string, Array<Run<PlayersEmbedded>>> = new Map();
 let gameListPromise: Option<Promise<void>> = null;
 let platformListPromise: Option<Promise<void>> = null;
 let regionListPromise: Option<Promise<void>> = null;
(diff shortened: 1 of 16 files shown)


wooferzfg:
I switched back to flexbox rather than using a table because it's much easier to work with for the mobile layout.

Also, we'll need to figure out how to get the Save button in the sidebar to update the splits table.

CryZe:
> We'll need to figure out how to get the Save button in the sidebar to update the splits table.

iirc, I implemented that. Is that not working anymore?

wooferzfg:
It only updates the table after you go to a different view and come back.

CryZe:
Oh I believe it only updates it when the splits are not in the table at all. But I guess we need it for when the game / category changed as well (might as well always update it then to make sure).

CryZe:
I rebased the changes onto master with the sidebar changes, and the sidebar updates everything properly now. Also this allowed moving lots of methods into the splits selection instead of the main file.

CryZe:
Looks like the tests don't properly load the splits anymore though.

wooferzfg:
I updated the tests to work with the new splits selection. We might still need to update some of the hashes, but we can do that closer to merging.

I had to put a wait time of half a second between importing a file and choosing which splits to open. We could potentially update this later if we end up adding some kind of spinner while the splits are being imported, but the rendering test's focus is on the layout rather than the UI, so I'm not too concerned either way.

CryZe:
This should mostly be good to go now I think. The race conditions can kind of still happen, but I was not really able to break anything when I played around with it. So I think we can fold this into the other issue of long term cleaning this up somewhat.

I also added support for dragging in splits files and I added a context menu to the export button. Also empty splits now show some default text.

Also urgh, what's going on with the CI again.

PR #301, opened 2020-04-11
Use actions-rs/install for wasm-gc and wasm-bindgen-cli

### LINKED ISSUE #300: Switch rust binaries in CI to this
https://svartalf.info/posts/2020-04-10-github-action-for-binary-crates-installation/

wooferzfg:
Fix #300

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,12 +59,17 @@ jobs:
         cd binaryen
         cmake .
         make wasm-opt
-        cargo install wasm-gc
 
-    - name: Install wasm-bindgen
-      run: |
-        set -e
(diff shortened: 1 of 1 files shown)


CryZe:
Mmh, seems like not much of an improvement without the tool-cache. So let's request them before merging this.

CryZe:
I believe they might reject wasm-gc, as the repository is "deprecated" in favor of using wasm-pack, which we may want to look into using. Though honestly it seems like a whole lot of unnecessary work to use wasm-pack.

CryZe:
Alright, turns out we don't need wasm-gc at all, as it's now (actually has been for a while) part of wasm-bindgen. And wasm-bindgen-cli has been added to the toolchain-cache.

PR #302, opened 2020-04-11
Update pngjs to 4.0.0

wooferzfg:
The update in https://github.com/LiveSplit/LiveSplitOne/pull/298 also bumped the pngjs version that a bunch of other packages use to 4.0.0, even though they don't support it.


CryZe:
Did you have to manually undo it, or is greenkeeper just bad?

wooferzfg:
greenkeeper is bad. I just ran `npm install pngjs@latest`

PR #304, opened 2020-04-14
Extract SidebarContent to each view

### LINKED ISSUE #303: Refactor the Sidebar to be part of the view
Each view should have its own sidebar. We shouldn't have one mega sidebar that is outside of each view, because communicating with that sidebar is non-trivial.

wooferzfg:
Fix #303


wooferzfg:
I switched to `execSync` in `buildCore.js` because `spawnSync` doesn't give any output when a command fails to start a process.

wooferzfg:
Seems like `wasm-bindgen` isn't being found: https://github.com/LiveSplit/LiveSplitOne/pull/304/checks?check_run_id=584608702#step:15:206

Not sure if we're doing something wrong with `actions-rs/install`.

CryZe:
Ah, it's a new two week cycle. Seems like the tool-cache never worked, we just still had it cached. It seems like it's called wasm-bindgen-cli in the tool-cache, not wasm-bindgen. Though that should be the same binary.

CryZe:
I filed an issue here: https://github.com/actions-rs/tool-cache/issues/4

CryZe:
I switched back to manually installing wasm-bindgen, so that we can merge this.

PR #328, opened 2020-04-19
Separate the Timer and Layout views into separate components

wooferzfg:


CryZe:
It seems like I can't switch between splits and layouts by dragging them in anymore.

wooferzfg:
Should be fixed now. We were losing the `this` context on the split and layout opening functions before.

CryZe:
Other than the CI failure, it seems fine

PR #347, opened 2020-04-29
Replace checkboxes with toggles

### LINKED ISSUE #36: Checkboxes stretched out in layout editor
screenshot

wooferzfg:
Fix #36


wooferzfg:
I think that jumpy behavior was there for the checkboxes as well, but yeah I'll fix that + the other issues you mentioned.

CryZe:
Oh and can we make them slightly larger too? Maybe just that single pixel or so.

wooferzfg:
I switched to a grid layout for our `Settings` component, which makes things a lot more consistent there. I also cleaned up some of `Table.scss`, which resulted in the Run Editor's icon column no longer being weirdly different from the rest of the columns.

PR #363, opened 2020-05-12
Update how splits are loaded from the leaderboard

### LINKED ISSUE #362: Downloading splits from the Leaderboard should act as new splits
If you download splits from the leaderboard in the Run Editor, it'll override the run that is getting modified by the run editor. However if you save that, there's two possible suboptimal behaviors:

1. If you edited the splits that you had open in your timer, it'll correctly replace the splits in the timer. However it will not reset your splits key. So the splits will still be associated to your splits and those will be overridden once you save. The downloaded splits should be regarded as fresh splits (splits key == undefined).

2. If you edit splits in your database and exit the run editor with the downloaded splits, then the splits in the database will be overridden, which you likely didn't intend to do. We should create a new entry in the database.

wooferzfg:
- The Run Editor now shows a loading spinner while it's loading splits from the leaderboard.
- I added a new property to the Run Editor menu state, `persistChanges`. This determines whether changes made in the Run Editor should immediately be saved to the database or not. We were previously using the splits key to determine this, but we now need that in order to switch between runs while the Run Editor is open.

Fix #362


wooferzfg:
@CryZe What do we want the behavior to be in this situation?

5dSeawif1t

Currently, I'm giving the imported splits an `undefined` splits key, which means it doesn't automatically get persisted to the database.

src/ui/RunEditor.tsx
```diff
@@ -1858,6 +1872,18 @@ export class RunEditor extends React.Component<Props, State> {
     }
 
     private async downloadSplits<T>(apiRun: Run<T>, apiUri: string) {
+        // TODO: Determine whether there are any unsaved changes in the Run Editor
```
wooferzfg, line 1875: I think being able to close the Run Editor without disposing it would allow us to handle this?

src/ui/TimerView.tsx
```diff
@@ -246,7 +246,7 @@ export class TimerView extends React.Component<Props, State> {
             toast.error(e);
         };
         this.connection.onmessage = (e) => {
-            // FIXME: Clone the Shared Timer. This assumes that `this` is always
+            // TODO: Clone the Shared Timer. This assumes that `this` is always
```
CryZe, line 249: I started using FIXME as long term things that need to be fixed and TODOs being temporary self reminders to do something as part of the ongoing PR. So unless you plan to fix this as part of this PR, this should stay a FIXME.

wooferzfg: In that case, can we rename our existing TODOs to FIXMEs instead?

CryZe:
If you edit splits that aren't open and download splits, it currently replaces the splits in your timer, despite you never "touching your timer". I don't think that's what we want. Honestly, I feel like we long term probably don't even want to have the leaderboard in the run editor. But that's outside the scope of this PR.

wooferzfg:
@CryZe Should we always persist the splits from the leaderboard to the database if you click "OK"? That seems like the least confusing user experience based on what you said.

CryZe:
Yeah I think that may be fine? I'm honestly not sure.

wooferzfg:
Also, I agree that the leaderboard shouldn't be in the Run Editor.

wooferzfg:
@CryZe Can we merge this? I don't think there are remaining issues with this, at least as a fix for #362.

CryZe:
I can re-review this tomorrow maybe.

wooferzfg:
@CryZe Could you take a look at the way I'm handling the run editor in this PR? Not sure if I'm disposing it properly everywhere

CryZe:
I'll close this, because splits.io will close down soon anyway.

PR #365, opened 2020-05-16
Only re-render components when necessary

wooferzfg:
Using `shouldComponentUpdate` allows us to skip updates unless the layout state actually changes. I'm not sure if we want to write our own comparison function rather than relying on `fast-deep-equal`, especially since our layout state values are guaranteed to be JSON. We could also just rely on `livesplit-core` to tell us whether a component needs to be updated.

### Benchmarks

Total scripting time when timer isn't running **(before)**: 19.6%
Total scripting time when timer isn't running **(after)**: 9.6%

Total scripting time when timer is running **(before)**: 21.5%
Total scripting time when timer is running **(after)**: 12.4%

[benchmarks.zip](https://github.com/LiveSplit/LiveSplitOne/files/4638639/benchmarks.zip)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2811,6 +2811,14 @@
         "fast-json-stable-stringify": "^2.0.0",
         "json-schema-traverse": "^0.4.1",
         "uri-js": "^4.2.2"
+      },
+      "dependencies": {
+        "fast-deep-equal": {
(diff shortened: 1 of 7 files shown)


CryZe:
Yeah looks like a decently big improvement on my side as well. This is great :D

PR #378, opened 2020-06-06
Improve Paint Times

CryZe:
In most frames only the timer will be redrawn. By moving it into its own browser layer, we can reduce the amount of work the browser needs to do to redraw that layer.

You can read more about this here:

https://developers.google.com/web/fundamentals/performance/rendering/simplify-paint-complexity-and-reduce-paint-areas#promote_elements_that_move_or_fade

This improves our paint times from 0.71ms

to about 60us

### DIFF
--- a/src/css/Timer.scss
+++ b/src/css/Timer.scss
@@ -1,0 +1,4 @@
+.timer {
+    will-change: transform;
+    transform: translateZ(0);
+}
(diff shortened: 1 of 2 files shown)


CryZe:
I split out this PR from the other changes as this is a clear win, while the other changes don't seem to improve much (and possibly even regress the performance?!)

PR #380, opened 2020-06-11
Add parent params to twitch embeds

### LINKED ISSUE #379: Twitch Embeds don't work anymore
https://discuss.dev.twitch.tv/t/twitch-embedded-player-updates-in-2020/23956

wooferzfg:
I'm not sure how we can test this on one.livesplit.org without just deploying these changes.

Fix #379

### DIFF
--- a/src/ui/Embed.tsx
+++ b/src/ui/Embed.tsx
@@ -60,5 +60,6 @@ function resolveYoutube(videoId: string): JSX.Element {
 }
 
 function resolveTwitch(videoId: string): JSX.Element {
-    return videoIframe(`https://player.twitch.tv/?video=${videoId}&autoplay=false`);
+    const domain = window.location.hostname;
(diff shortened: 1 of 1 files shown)


src/ui/Embed.tsx
```diff
@@ -60,5 +60,6 @@ function resolveYoutube(videoId: string): JSX.Element {
 }
 
 function resolveTwitch(videoId: string): JSX.Element {
-    return videoIframe(`https://player.twitch.tv/?video=${videoId}&autoplay=false`);
+    const domain = process.env.NODE_ENV === "production" ? "one.livesplit.org" : "localhost";
```
CryZe, line 63: In the thread I linked the dynamically get the domain from the browser. Maybe that makes more sense than hardcoding the values.

wooferzfg: Seems like `window.location.hostname` works:

chrome_JCWsWMx0qg
chrome_YWuXNR5PP6

PR #381, opened 2020-06-14
Apply all the recent performance improvements

CryZe:
This updates livesplit-core which brings a variety of performance improvements:

- The Layout State is now being reused and thus most frames don't require any heap allocations anymore. However we still serialize everything over into a JSON string for now, which puts a lot of
  garbage on the JS heap.
  https://github.com/LiveSplit/livesplit-core/pull/334

- The frequent performance.now() calls we do, first look up the window and performance object every time. This tripled the amount of calls we do over into JavaScript, with each call into JavaScript being quite expensive in Chrome.
  https://github.com/LiveSplit/livesplit-core/pull/335

- By introducing a timer snapshot mechanism we further reduce the calls to performance.now() to a single time.
  https://github.com/LiveSplit/livesplit-core/pull/339

- Rust 1.44 regressed the performance of 128-bit integer multiplications by accident. Those are used for hashing the comparisons when looking up the times for a comparison, which is something we do very frequently. We however don't have many comparisons, so a simple Vec that we loop through is a bit faster, even in native code, and quite a bit faster in the web, because of the Rust 1.44 regression.
  https://github.com/LiveSplit/livesplit-core/pull/338

- We delay registering the Gamepad Hook Interval until the first gamepad button is registered. Most people won't use a gamepad, so the interval just wastes cpu time for no reason.
  https://github.com/LiveSplit/livesplit-core/pull/340

### DIFF
--- a/src/ui/LiveSplit.tsx
+++ b/src/ui/LiveSplit.tsx
@@ -3,8 +3,7 @@ import Sidebar from "react-sidebar";
 import {
     HotkeySystem, Layout, LayoutEditor, Run, RunEditor,
     Segment, SharedTimer, Timer, TimerRef, TimerRefMut,
-    HotkeyConfig,
-    LayoutState,
+    HotkeyConfig, LayoutState,
 } from "../livesplit-core";
 import { convertFileToArrayBuffer, convertFileToString, exportFile, openFileAsString } from "../util/FileUtil";
 import { Option, assertNull, expect, maybeDisposeAndThen, panic } from "../util/OptionUtil";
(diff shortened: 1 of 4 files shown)


src/ui/LiveSplit.tsx
```diff
@@ -4,6 +4,7 @@ import {
     HotkeySystem, Layout, LayoutEditor, Run, RunEditor,
     Segment, SharedTimer, Timer, TimerRef, TimerRefMut,
     HotkeyConfig,
+    LayoutState,
```
wooferzfg, line 7:
```suggestion
    HotkeyConfig, LayoutState,
```

PR #397, opened 2020-07-27
Align web socket commands with LiveView server

etopiei:
Currently the web socket commands are not named the same way as those used by the LiveSplit server component.
I have written a script to connect the Android Remote app to LiveSplitOne ([here](https://github.com/etopiei/livesplit-remote-forwarder)) but as some of the commands are different and/or not supplied it is tricky to get this working. In order for more tools to be able to interface better with LiveSplitOne I think it would be good to have standardized commands that are the same as the desktop app.

I am fairly new to this code base, so let me know if anything needs to change or could be done differently, I'm open to feedback and different ideas of how this should be done. This is currently more of a POC, but I'd be happy to help bring this to LiveSplitOne.


src/ui/TimerView.tsx
```diff
@@ -252,13 +253,16 @@ export class TimerView extends React.Component<Props, State> {
...
+                    case "starttimer": this.start(); break;
+                    case "getcurrenttime": this.getTime(); break;
+                    case "getcurrenttimerphase": this.getTimerPhase(); break;
                     case "split": this.split(); break;
                     case "splitorstart": this.splitOrStart(); break;
                     case "reset": this.reset(); break;
-                    case "togglepause": this.togglePauseOrStart(); break;
-                    case "undo": this.undoSplit(); break;
-                    case "skip": this.skipSplit(); break;
+                    case "pause": this.togglePauseOrStart(); break;
```
CryZe, line 262: I'd prefer this to either stay `togglepause` or for `pause` to be a (deprecated) alias.

CryZe:
Considering there's all sorts of new functions dedicated entirely to the WebSocket communication, I'd probably also would want to see all of this logic moved out into a new source file.

src/ui/TimerView.tsx
etopiei: Yeah that's a good idea, I'll fix that up :+1:

etopiei:
@CryZe  I've re-worked this a bit as you suggested, hopefully this is looking a bit better now. Let me know if you'd like any other changes or would like to try things another way. Cheers :+1:

src/ui/TimerView.tsx
```diff
@@ -353,4 +368,4 @@ export class TimerView extends React.Component<Props, State> {
     private resumeGameTime() {
         this.writeWith((t) => t.resumeGameTime());
     }
-}
+}
```
wooferzfg, line 371: Can we make sure that there are newlines at the end of files?

src/util/WebSocket.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function formatTimeForServer(totalSeconds: number): string {
+    const t = formatLeaderboardTime(totalSeconds, false);
+    const parts = t.split(":");
+    if (parts[0].length < 2) {
+        return "0" + parts[0] + ":" + parts[1].slice(0, parts[1].length - 1);
+    } else {
+        return parts[0] + ":" + parts[1].slice(0, parts[1].length - 1);
+    }
+}
```
wooferzfg, line 20: Same here.

src/util/WebSocket.ts
```diff
@@ -0,0 +1,20 @@
+import { formatLeaderboardTime } from './TimeUtil';
+
+export function mapPhaseEnumToString(phase: number) : string {
+    const phases = ["NotRunning", "Running", "Ended", "Paused"];
+    if (phase > 0 && phase < phases.length) {
+        return phases[phase];
+    } else {
+        return "";
+    }
+}
```
wooferzfg, line 10: Couldn't we just do:
```
return phases[phase] || "";
```

src/util/WebSocket.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function formatTimeForServer(totalSeconds: number): string {
+    const t = formatLeaderboardTime(totalSeconds, false);
+    const parts = t.split(":");
+    if (parts[0].length < 2) {
+        return "0" + parts[0] + ":" + parts[1].slice(0, parts[1].length - 1);
+    } else {
+        return parts[0] + ":" + parts[1].slice(0, parts[1].length - 1);
+    }
```
wooferzfg, line 19: This method seems really hacky. Could we either modify `formatLeaderboardTime` to take additional arguments or just add an additional method to `TimeUtil`?

src/util/WebSocket.ts
CryZe: Ideally we should use `expect(phases[phase], "Unknown timer phase supplied")` here or so, as the index out of bounds case should visibly explode and not implicitly do something weird.

CryZe:
While you've moved some functionality over into WebSocket.ts, I'd actually like to see *all* the WebSocket logic be moved over. There should be close to nothing left in the TimerView.

etopiei:
@CryZe  Sorry this has taken so long, I've been a bit busy lately, but I've created a new WebSocket manager class which handles the web socket setup and communication. I think this is looking a bit cleaner now, but let me know if there are any further changes required. :+1:

CryZe:
The protocol changed a lot since then, so I'll close this.

Issue #493, opened 2021-01-26
Private Browsing on Firefox disables IndexedDB


Rektroth:
I have tested using LiveSplitOne on my chromium browser, and it worked fine on there.

On Firefox, LiveSplitOne gets stuck on the loading screen. A quick look at the console reveals the following message:
```
Uncaught (in promise) DOMException: A mutation operation was attempted on a database that did not allow mutations. bundle.js:1:62182
  c           https://one.livesplit.org/bundle.js:1
  (Async: promise callback)
  b           https://one.livesplit.org/bundle.js:1
  a           https://one.livesplit.org/bundle.js:1
  (Async: promise callback)
  b           https://one.livesplit.org/bundle.js:1
  a           https://one.livesplit.org/bundle.js:1
  (Async: promise callback)
  b           https://one.livesplit.org/bundle.js:1
  a           https://one.livesplit.org/bundle.js:1
  (Async: promise callback)
  b           https://one.livesplit.org/bundle.js:1
  a           https://one.livesplit.org/bundle.js:1
  (Async: promise callback)
  b           https://one.livesplit.org/bundle.js:1
  l           https://one.livesplit.org/bundle.js:1
  l           https://one.livesplit.org/bundle.js:1
  <anonymous> https://one.livesplit.org/bundle.js:1
  <anonymous> https://one.livesplit.org/bundle.js:1
  a           https://one.livesplit.org/bundle.js:1
  <anonymous> https://one.livesplit.org/bundle.js:1
  a           https://one.livesplit.org/bundle.js:1
  <anonymous> https://one.livesplit.org/bundle.js:1
  <anonymous> https://one.livesplit.org/bundle.js:1
```

CryZe:
It seems like IndexedDB is not usable at all when you are using Firefox's private browsing.

CryZe:
Apparently they consider this a bug in Firefox and are currently intending to fix this: https://bugzilla.mozilla.org/show_bug.cgi?id=1639542

qwertychouskie:
Looks like this should be fixed in the upcoming Firefox 115

CryZe:
Yeah this seems to be fixed.

PR #496, opened 2021-01-27
Alert for #493

Rektroth:
IndexedDB is not usable at all when you are using Firefox's private browsing. Apparently they consider this a bug in Firefox and are currently intending to fix this: https://bugzilla.mozilla.org/show_bug.cgi?id=1639542
However, some discussion seems to indicate that it may be intentional and not be fixed: https://bugzilla.mozilla.org/show_bug.cgi?id=781982
Until (and if) they fix it, LiveSplitOne can display an alert telling users to turn off private browsing.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -56,8 +56,9 @@ Those are known to block WebAssembly.`);
         if (e.name == "InvalidStateError") {
             alert(`Couldn't load LiveSplit One. \
 You may be in private browsing mode. \
-LiveSplitOne cannot run in this mode. \
-To run LiveSplitOne, please disable private browsing in your settings.\n`);
+LiveSplit One cannot store any splits, layouts, or other settings because of the limitations of the browser's private browsing mode. \
+These limitations may be lifted in the future. \
+To run LiveSplit One for now, please disable private browsing in your settings.\n`);
         }
     }
 }
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -16,40 +16,49 @@ async function run() {
     const ReactDOM = await ReactDOMImport;
     const { toast, ToastContainer } = await Toastify;
 
-    const {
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -16,40 +16,49 @@ async function run() {
...
 You may be using a browser that doesn't support WebAssembly. \
 Alternatively, you may be using an Adblocker like uBlock Origin. \
 Those are known to block WebAssembly.`);
+        }
+    } catch (e) {
+        if (e.name == "InvalidStateError") {
+            alert(`Couldn't load LiveSplit One. \
+You may be in private browsing mode. \
+LiveSplitOne cannot run in this mode. \
+To run LiveSplitOne, please disable private browsing in your settings.\n`);
```
CryZe, line 60:
```suggestion
You may be in private browsing mode. \
LiveSplit One cannot store any splits, layouts, or other settings because of the limitations of the browser's private browsing mode. \
These limitations may be lifted in the future. \
To run LiveSplit One for now, please disable private browsing in your settings.\n`);
```

src/index.tsx
```diff
@@ -16,40 +16,49 @@ async function run() {
...
+                document.getElementById("base"),
+            );
+        } catch (_) {
+            alert(`Couldn't load LiveSplit One. \
 You may be using a browser that doesn't support WebAssembly. \
 Alternatively, you may be using an Adblocker like uBlock Origin. \
 Those are known to block WebAssembly.`);
+        }
+    } catch (e) {
+        if (e.name == "InvalidStateError") {
```
CryZe, line 56: /home/runner/work/LiveSplitOne/LiveSplitOne/src/index.tsx:56:20
ERROR: 56:20  triple-equals  == should be ===

PR #569, opened 2021-07-01
Implement Support for Custom Fonts

CryZe:
This implements support for custom fonts. Additionally it updates livesplit-core and all our npm dependencies to the latest version. This also switches how we propagate colors and fonts through the layout. We are now using custom CSS variables instead of passing them through the component tree.


CryZe:
@wooferzfg I'll merge it now, there's a few follow-up PRs I'd like to look into. I can sneak any changes you'd like to see into those.

PR #581, opened 2021-07-24
Use `make -j2` instead of `make -j`

CryZe:
Apparently `-j` completely crashes the GitHub Actions runner (or at least makes it hang).


CryZe:
As far as I can tell, we don't even do a release build, so it's a bit absurd that building like 4 C files or whatever in parallel uses that much RAM.

PR #666, opened 2022-01-19
Updated regular Layout to use livesplit-core software rendering

Hurricane996:


CryZe:
We switched to canvas.

PR #716, opened 2022-09-28
Update `livesplit-core`

CryZe:
This updates `livesplit-core` to the latest version. This mostly adds custom variable columns and allows for the timer to visualize the delta for the background. There's some performance improvements as well.

Also we now build `livesplit-core` in debug mode by default when building locally. So compilation is a little bit faster and additional debug checks are active, such as integer overflows.


wooferzfg:
I think the CI issue is because our version of `commonmark` is too new for `commonmark-react-renderer`, which isn't maintained anymore. We could probably use `react-markdown` instead: https://github.com/remarkjs/react-markdown

CryZe:
How is this even a thing though? Why is it ignoring our lockfile?

wooferzfg:
For now, we could downgrade our version of `commonmark`, or add `--legacy-peer-deps` to the `npm install` command in CI.

PR #765, opened 2023-01-24
Implement more Detailed Timer Settings

CryZe:
This implements the following settings / features for the Detailed Timer:
1. The ability to set / override the colors of the timer, segment timer, segment name, comparison names and comparison times.
2. The ability to turn off the gradient of the timer and segment timer.
3. The ability to set the accuracy of the comparison times.

This also updates all the npm packages to their latest versions.


CryZe:
Since this bumps `livesplit-core`, various other changes get pulled in as well, such as the faster extraction of PNG dimensions which resulted in 30% faster parsing times.

PR #874, opened 2024-05-09
Add Tooltips

CryZe:
This adds tooltips to all settings. It's not optimal yet, because it doesn't support mobile yet, where you don't have a cursor that could hover the settings.


wooferzfg:
I'm actually able to view the tooltips on mobile by tapping on the label text, but it's currently cut off on the left side

Screenshot_20240509_113942_Chrome Beta

CryZe:
Interesting. I wonder if that's a specific browser behavior. Feel free to improve it from here.

PR #875, opened 2024-05-09
Fix tooltip positions on mobile

wooferzfg:
Tooltips can be opened on mobile (at least in Chrome on Android) by tapping on the label.

### DIFF
--- a/src/css/Tooltip.scss
+++ b/src/css/Tooltip.scss
@@ -1,3 +1,4 @@
+@import 'mobile';
 @import 'variables';
 
 .tooltip {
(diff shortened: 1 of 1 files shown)


src/css/Tooltip.scss
```diff
@@ -22,6 +23,11 @@
     opacity: 0;
     transition: opacity 0.25s;
     transition-delay: 0.5s;
+
+    @include mobile {
+        left: 0;
+        margin-left: 0;
+    }
```
CryZe, line 30: I'm not sure this works everywhere? For example we have the tooltips in the splits editor variables tab as well.

wooferzfg: I tested those and it seemed fine

Issue #888, opened 2024-05-26
Show the latest updates in the About page


CryZe:
Unlike the original LiveSplit, the web version just updates automatically in the background. This means that unless you keep up with the repository, you don't get to know about new features. I think we should query the most recently merged PRs that are tagged in a particular way to build a list of the most recent updates as part of the about page (or possibly somewhere else).

wooferzfg:
I also noticed that with a service worker, it usually takes an additional refresh to get the latest version. Something I built into one of my projects with a service worker is a [prompt about there being a new update](https://github.com/wooferzfg/tww-rando-tracker/blob/main/src/ui/launcher.jsx#L60), and if you click on the prompt it refreshes the page. Maybe we want something similar for LSO?

CryZe:
If that's easily possible, then yeah we should totally have that. The additional refresh is somewhat intended, as that means the website will load much quicker and the latest features aren't usually all that important, but if we can have a little toast popup, then that would be great.

PR #889, opened 2024-05-26
Prompt when there is an update

wooferzfg:
https://github.com/LiveSplit/LiveSplitOne/issues/888#issuecomment-2132244331

When opening LiveSplit One, the service worker downloads the latest files in the background, which usually takes a few seconds. After the files are downloaded, if it detects a change, it will now show a popup message in the bottom-right corner that can be clicked to auto-refresh the page. (The message will automatically disappear after 5 seconds to avoid being too intrusive.)

https://github.com/LiveSplit/LiveSplitOne/assets/8262173/cad287d6-1ed7-48a0-9318-b86211bb9b76


CryZe:
Seems to be red in CI (eslint), otherwise looks good.

PR #895, opened 2024-05-27
Clean up and move server connection to settings

CryZe:
This moves the server connection button to the settings and cleans the entire functionality up in various ways. The server connection no longer gets closed by navigating around the application. Additionally the server URL is now being remembered between sessions.

This also improves the design of the toast messages.

image


wooferzfg:
Wow, the toast messages look great

PR #898, opened 2024-05-31
Add a changelog and improve contributors section

CryZe:
This improves the contributors section in the About page. It now uses flexbox to dynamically place the contributors. Additionally their avatar is now being shown.

Also the ten most recent changes are now being shown in the About page. The information is extracted from the git history. The idea is to mark commit messages with a special changelog section at the end that's specifically meant to be shown as a user facing change. It can be arbitrary markdown and looks like this (which is also the first one):

Changelog: The About page now shows the most recent changes. The contributors section has also been improved. It now shows the avatar of each contributor.

### DIFF
--- a/src/ui/About.tsx
+++ b/src/ui/About.tsx
@@ -1,6 +1,8 @@
 import * as React from "react";
 
 import LiveSplitIcon from "../assets/icon.svg";
+import { renderMarkdown } from "../util/Markdown";
+import variables from "../css/variables.scss";
 
 import "../css/About.scss";
 
@@ -13,6 +15,13 @@ interface Callbacks {
     openTimerView(): void,
 }
 
+const changelogRenderSettings = {
+    softBreak: false,
+    escapeHtml: false,
+};
+
+const contributorAvatarSize = parseFloat(variables.contributorAvatarSize);
+
 export class About extends React.Component<Props> {
     public render() {
         const renderedView = this.renderView();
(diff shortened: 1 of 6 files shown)


src/ui/About.tsx
```diff
@@ -21,6 +27,8 @@ export class About extends React.Component<Props> {
     }
 
     private renderView() {
+        const idealAvatarResolution = Math.round(devicePixelRatio * 42);
```
wooferzfg, line 30: Can we export this 42 from the SCSS and import it here, like we do in `variables.scss`?

CryZe: Yeah, though I wasn't fully sure how to do this? What do I do with the "px" part?

wooferzfg: I think it just works with `parseFloat` 
https://github.com/LiveSplit/LiveSplitOne/blob/909d61568d244fab5de77045085b845232a59fbb/src/css/variables.scss#L36
https://github.com/LiveSplit/LiveSplitOne/blob/909d61568d244fab5de77045085b845232a59fbb/src/ui/LiveSplit.tsx#L30

CryZe: Oh, similar code uses `parseFloat`, which just silently gets rid of it.

wooferzfg: Maybe we can add a comment that explains that behavior

CryZe: Forgot about the comment, but I did fix it.

PR #917, opened 2024-06-08
Skip update prompts in invalid situations

### LINKED ISSUE #913: Initial install of service worker triggers "update found" message
Apparently `controllerchange` is not quite the right event. There's `updatefound` which might be the better way:

https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration#examples

cc @wooferzfg

wooferzfg:
When it's the first install of the service worker, or when the user clears the site data, the `controllerchange` event gets triggered. However, in these situations, we want to avoid prompting the user about an update.

Resolves #913


CryZe:
Thanks :)

PR #921, opened 2024-06-08
Fix the margin for the PWA Icon on iOS

CryZe:
This adjusts the margin for the PWA Icon on iOS to roughly match the Apple guidelines, but more importantly, to just finally look good.

image

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -120,6 +120,10 @@ export default async (env, argv) => {
                     theme_color: "#232323",
                     appleStatusBarStyle: "black-translucent",
                     icons: {
+                        appleIcon: {
+                            offset: 10,
+                        },
+                        windows: false,
                         coast: false,
                         yandex: false,
                     },


CryZe:
@wooferzfg Does the PWA on Android have similar problems? (See the linked issue as well)

wooferzfg:
It looks fine on Android: 

Screenshot_20240608_120234_One UI Home

CryZe:
I'm not sure I like that the background is white there, but that's a separate issue.

wooferzfg:
It's possible that it looks different on other Android phones, though. I use a Samsung phone, which has rounded rectangles as icons, but other Android phones have circular icons.

wooferzfg:
This is what it looks like on a Pixel phone, seems fine as well:

Screenshot_20240608_121037_Signal

CryZe:
image

This is how it looks like on iOS now (with similar looking icons beside it).

PR #951, opened 2024-07-21
Automatically release Tauri binaries

CryZe:
This makes it so we automatically release the Tauri binaries when we merge a new change to the master branch.


.github/workflows/ci.yml
```diff
@@ -195,3 +200,20 @@ jobs:
...
+          TARGET: ${{ matrix.binaryen }}
+
+      - name: Release (Tauri)
+        if: matrix.platform == 'tauri'
+        uses: softprops/action-gh-release@v2
+        with:
+          files: LiveSplitOne-*.*
+          name: Latest
+          tag_name: latest
+          body: This contains the latest version of the auto splitter.
```
wooferzfg, line 219: Is this supposed to say "auto splitter"?

CryZe: Oh, no, I just copied it from our auto splitter template. But as with all things CI, this will fail a couple of times anyway, so I didn't pay too much attention to minor stuff like this yet, but yeah this needs to be fixed.

PR #1026, opened 2025-03-24
Switch to Lucide Icons

CryZe:
We have had a really outdated version of Font Awesome directly vendored into our repository. Not only was it outdated, but it also is shipped as a font file, including all the icons, even though we only use a small subset of them. By switching to Lucide, we not only are depending on something more up to date, but now the icons can also be tree shaked.

This also updates some of our CSS to use flex box and grid instead of manually positioning with margins.


CryZe:
As expected, this shaved off around 10% of our compressed bundle size.

PR #1110, opened 2026-01-12
Fixed reset not considering the message's argument

Paloys:
Before this pull request, LiveSplitOne would not consider the server's "saveAttempt" argument when receiving a reset command. This makes sure the command is full taken into account, by asking the user if they wish to save the times if and only if the command has not provided the "saveAttempt" argument.


Paloys:
The bindings at https://github.com/LiveSplit/livesplit-core/blob/a135301b008d9211ae37ae14b4dc6cec5a3c2aaa/capi/src/web_command_sink.rs#L144 show that the option is passed to the TypeScript bindings, but LiveSplitOne didn't consider it.

PR #1137, opened 2026-04-27
fix(tauri): Fix Linux Tauri Popout Windows

### LINKED ISSUE #1106: [LINUX] LiveSplit not popping out
when i press the pop out button nothing happens.

DaVinciLord:
Closes #1106

Handle WebKitGTK window.open requests natively so popout windows work in the Tauri app on Linux. 
Add a localized error message for cases where no popout window can be created.

### DIFF
--- a/src-tauri/src/constants.rs
+++ b/src-tauri/src/constants.rs
@@ -1,0 +1,7 @@
+pub const MAIN_WINDOW_LABEL: &str = "main";
+pub const POPOUT_WINDOW_LABEL_PREFIX: &str = "popout";
+pub const APP_TITLE: &str = "LiveSplit One";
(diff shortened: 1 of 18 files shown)


IvarWithoutBones:
I've been using this for a few months now and it works really well! Nice to see Linux support caught up with other platforms.

I noticed one minor issue though: when closing the main window while the popout is open, the popout stays open. The content does get cleared so you're left with a mostly blank window, not a huge deal but I figured I might as well mention it.

DaVinciLord:
Thanks for using this modified version ! I will soon circle back to this (when I regain access to my Linux rig) as I see that the branch is out of date and will also address the issue you mentioned.

Feel free to report any issues that falls in the same scope. :)

